PR #13, opened 2026-05-23
refactor(log): route all engine output through EngineLog

Delarkz:
## Summary

Every diagnostic line the engine emits now goes through one path (`EngineLog.emit`) with one sink contract: OSLog always, optional host handler. The stdio fallback that previously caused duplicate lines in Xcode is gone, and the SW playback subsystem gets its own log category (`sw.playback`) so it can be isolated from the rest of the engine in `log stream` / Console.app filters.

## Motivation

Two distinct problems with the old logging layout:

1. **Duplicate lines in Xcode.** `EngineLog.emit` wrote to OSLog unconditionally, then *also* called `print(line)` whenever no host handler was installed. Xcode renders both OSLog activity and the process's stdout in the same debug-area console, so every emit surfaced twice. The old doc comment claimed gating handler vs. stdout solved this — it did not, because OSLog still fired in parallel. A `isatty(STDOUT_FILENO)` gate was tried as an intermediate fix; Xcode wires GUI-app stdout to a pty, so the gate let duplicates through anyway. The fallback was removed entirely.

2. **23 raw `print(...)` calls** scattered across `AVIOReader`, `Demuxer`, `AudioDecoder`, `AudioOutput`, `SoftwareVideoDecoder`, `SampleBufferRenderer`, plus a vestigial duplicate in `AetherEngine` itself. These bypassed OSLog entirely, so they weren't filterable by category, didn't reach Console.app, and contributed to the Xcode duplication when their host also emitted through `EngineLog`.

The new `swPlayback` category carves the custom playback subsystem (`SoftwarePlaybackHost` + its decoders + `SampleBufferRenderer` + `AudioDecoder` + `AudioOutput`) out of the `.engine` catch-all, so `.engine` now genuinely means "cross-subsystem / lifecycle" and the SW route is filterable on its own.

## Behaviour matrix

| Context | OSLog | Handler called | Net effect |
|---|---|---|---|
| Host app in Xcode, no handler installed | ✓ | — | one line per emit in Xcode console (was two) |
| Host app on device | ✓ | — | visible via Console.app / `log stream`, no stdio noise |
| `aetherctl serve` / `validate` (tty / piped / redirected) | ✓ | ✓ | timestamped stdout, identical to before |
| `aetherctl probe` (tty / piped / redirected) | ✓ | ✓ (newly installed) | identical to pre-change stdout output |
| Any future host with no handler | ✓ | — | OSLog only; install a handler to mirror to stdout |

### DIFF
--- a/Sources/AetherEngine/AetherEngine.swift
+++ b/Sources/AetherEngine/AetherEngine.swift
@@ -1680,14 +1680,6 @@ public final class AetherEngine: ObservableObject {
                     + "avBufBehind=\(String(format: "%.1f", bufferBehindSec))s"
 
                 EngineLog.emit(line, category: .engine)
-                // Also write to stdout so the line shows up in Xcode's
-                // debug console even on devices where the OSLog stream
-                // isn't being captured. EngineLog's own stdout fallback
-                // is suppressed once Sodalite installs its LogTap
-                // handler, so we duplicate the memprobe line here only.
-                // One line per 30 s is cheap; the rest of EngineLog stays
-                // single-sink to avoid console spam.
-                print(line)
             }
         }
     }
--- a/Sources/AetherEngine/Audio/AudioDecoder.swift
+++ b/Sources/AetherEngine/Audio/AudioDecoder.swift
@@ -85,7 +85,7 @@ final class AudioDecoder: @unchecked Sendable {
         // fully resolved layout/rate/format; initialise from it.
 
         #if DEBUG
-        print("[AudioDecoder] Opened: \(sampleRate)Hz, \(channels)ch, codec=\(String(cString: codec.pointee.name))")
+        EngineLog.emit("[AudioDecoder] Opened: \(sampleRate)Hz, \(channels)ch, codec=\(String(cString: codec.pointee.name))", category: .swPlayback)
         #endif
     }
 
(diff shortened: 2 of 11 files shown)


superuser404notfound (approved):
LGTM. Verified:

- Engine module is now 100% `print`-free (grep across `Sources/AetherEngine/` returns nothing); every diagnostic line flows through `EngineLog.emit`.
- Root cause of the Xcode duplicates was indeed `OSLog + print` firing in parallel, not the `print` fallback alone. Removing the stdio fallback in `EngineLog.emit` is the right fix.
- `aetherctl probe` now installs its own handler, which it needs because the previous behaviour relied on the now-removed stdio fallback. Behaviour matrix in the PR description matches what I see in `Sources/aetherctl/main.swift` (serve/validate/probe all install a handler now).
- New `.swPlayback` category cleanly carves the SoftwarePlaybackHost subsystem (`HardwareVideoDecoder`, `SoftwareVideoDecoder`, `SampleBufferRenderer`, `AudioDecoder`, `AudioOutput`) out of the `.engine` catch-all. `HardwareVideoDecoder` lives under `swPlayback` because it runs inside `SoftwarePlaybackHost`, not under the AVPlayer-driven path, so the name is accurate even though "Hardware" sounds like it might belong elsewhere.
- Public API unchanged: `handler` signature, `emit` signature, new `Category` case is non-breaking. `swift build` is green on this branch.

One follow-up I'll handle on the Sodalite side post-merge: `SodaliteApp.swift:18-19` has a stale comment ("App Store builds leave the handler nil so the engine emits to `print()` only") that no longer reflects reality after this change. Behaviour-wise the new world is better (stdout in App Store builds with no debugger attached was always wasted), but the comment will mislead future readers. Will fix in a separate commit once this lands.

Thanks for the cleanup.

PR #14, opened 2026-05-24
fix(muxer): drain libavformat's interleaver before each fragment cut

Delarkz:
## Summary

`MP4SegmentMuxer.cutFragmentForNextSegment` now drains libavformat's interleaver buffer (`av_interleaved_write_frame(ctx, nil)`) before triggering `mov_flush_fragment` (`av_write_frame(ctx, nil)`). Without the drain, packets still held in the interleaver — waiting for cross-stream DTS catch-up — were carried over into the *next* fragment's `moof+mdat` instead of landing in the one being cut.

## Motivation

`writePacket` uses `av_interleaved_write_frame`, which queues packets in libavformat's internal interleaver until enough cross-stream context is available to commit them in DTS order. For matroska sources where audio interleaves ahead of video (common for AC-3 / EAC-3 stream-copy), the interleaver buffers the trailing audio frames of the current segment waiting for the next IRAP to confirm ordering. When the producer triggers the segment cut, plain `av_write_frame(ctx, nil)` calls `mov_flush_fragment` directly — bypassing the interleaver. Anything still buffered re-emerges in the next fragment instead of finalising in the one being cut.

In practice this manifested as ~4 trailing AC-3 frames (~120 ms) consistently missing from the end of each segment's audio. The segment's declared `#EXTINF` therefore over-stated its audio coverage by the same amount, and the audio `tfdt` of every subsequent fragment fell behind the playlist's cumulative time.

## Fix

Insert `av_interleaved_write_frame(ctx, nil)` before each `av_write_frame(ctx, nil)` in `cutFragmentForNextSegment`. The interleaver drains its buffered packets into `mov_write_packet` first, populating the MOVTrack sample tables, and the subsequent `av_write_frame(ctx, nil)` then emits a `moof+mdat` that includes them.

## Verification

Tested against two sources with distinct codec mixes; mediastreamvalidator parses every segment cleanly in both runs.

**HEVC 10-bit 23.976 fps + AC-3 stereo, 343 segments.** Before the fix, each segment's audio fell short by ~4 AC-3 frames and the trailing frames spilled into the next segment, leaving every audio `tfdt` ~128 ms behind the playlist's cumulative time. After the fix, audio coverage matches `#EXTINF` within one AC-3 frame and tfdt is contiguous with the playlist:

| Segment | Audio range, before | Audio range, after |
|---|---|---|
| 0 | 309 frames, [0.000–9.888 s] | 313 frames, [0.000–10.016 s] |
| 1 | 313 frames, [9.888–19.904 s] | 313 frames, [10.016–20.032 s] |
| 2 | 313 frames, [19.904–29.920 s] | 313 frames, [20.032–30.048 s] |
| 5 | 312 frames, [49.952–59.936 s] | 312 frames, [50.080–60.064 s] |

Sample counts are mostly unchanged because the spillover preserved total frames across the stream — only seg-0's count moves. The fix corrects which segment each frame lands in, so the per-segment ranges shift forward by ~128 ms onto the playlist's grid.

**H.264 24 fps + EAC-3 5.1 Atmos JOC, 648 segments.** Every sampled segment has exactly 125 AC-3 frames × 1536 samples = 4.000 s of audio (the configured `targetDur`), tfdt strictly contiguous. mediastreamvalidator reports `Processed 648 out of 648 segments / Average segment duration: 3.998715`, no audio-coverage warnings.

### DIFF
--- a/Sources/AetherEngine/Video/MP4SegmentMuxer.swift
+++ b/Sources/AetherEngine/Video/MP4SegmentMuxer.swift
@@ -29,11 +29,16 @@ import Libavutil
 ///                          (which lives in fragments instead).
 ///   +default_base_moof  — relative offsets in tfhd (cleaner fmp4)
 ///   +frag_custom        — caller controls fragment cuts via
-///                          `av_write_frame(ctx, nil)`. The mp4
-///                          muxer's internal interleave queue holds
-///                          packets until cut time, naturally
-///                          aligning audio + video at fragment
-///                          boundaries.
+///                          `av_write_frame(ctx, nil)`. Packets enter
+///                          via `av_interleaved_write_frame` and
+///                          queue in libavformat's interleaver until
+///                          cross-stream DTS ordering allows commit
+///                          to `mov_write_packet`. At cut time we
+///                          must drain the interleaver explicitly
+///                          (see `cutFragmentForNextSegment`); the
+///                          cut itself bypasses the interleaver and
+///                          would otherwise leave still-buffered
+///                          packets to spill into the next fragment.
 ///   +delay_moov         — defers writing the moov atom until the
 ///                          first `av_write_frame(ctx, nil)` call,
 ///                          AFTER packets have been queued via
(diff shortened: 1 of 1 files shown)


superuser404notfound:
Merged as 88578b7. Sodalite pin bumped (d8ddfafd), ships with the next build.

Strong root-cause analysis. The per-segment tables make the impact obvious, and the JOC verification rules out collateral on the Atmos path. The kind of interleaver subtlety that only surfaces once you actually measure tfdt drift end-to-end across the full playlist.

PR #17, opened 2026-05-24
feat(diagnostics): bridge FFmpeg av_log into EngineLog .ffmpeg category

Delarkz:
## Summary

- New `FFmpegLogBridge` installs `av_log_set_callback` and forwards every line through `EngineLog.emit` under a new `.ffmpeg` category, formatted with `av_log_format_line2` so FFmpeg's own `[mp4 @ ...]` / `[h264 @ ...]` subsystem prefixes survive.
- New `.ffmpeg` case on `EngineLog.Category` keeps muxer / demuxer / decoder chatter filterable in Console.app independently of the engine's own log lines.
- `FFmpegLogBridge.install()` wired into `AetherEngine.init()` before any libav* entry point runs, so probe/load diagnostics land in Console.app + any host `EngineLog.handler` from the first call.

## Why

Without the bridge, FFmpeg writes to stderr via its default callback — invisible to App Store builds, in-app overlays, and `aetherctl`'s timestamped stdout. Real diagnostic value (corrupt headers, missing PTS, decoder errors) was being silently dropped.

## Behavior

- Default level: `AV_LOG_WARNING` — surfaces real problems without the per-segment `[mp4 @ ...] track ...` chatter `AV_LOG_INFO` emits during normal fMP4 muxing. Hosts wanting verbose output re-call `FFmpegLogBridge.install(level: AV_LOG_VERBOSE)`.
- Flags: `AV_LOG_PRINT_LEVEL` (severity survives into Console) + `AV_LOG_SKIP_REPEATED` (a decoder spamming the same warning collapses to one line + a `Last message repeated N times` follow-up).
- Output buffer: 1024 chars, matches ffmpeg's own default callback. Truncation acceptable for diagnostics; no re-alloc loop on overflow.
- Idempotent: re-calling `install()` overwrites the callback with the same function pointer, so the unconditional call in `AetherEngine.init()` is safe across multiple engine instances.
- Thread-safe by construction: the callback fires from libav* worker threads, and `EngineLog.emit` already promises thread-safety (OSLog is, and any host handler is required to be by `EngineLog`'s contract).

### DIFF
--- a/Sources/AetherEngine/AetherEngine.swift
+++ b/Sources/AetherEngine/AetherEngine.swift
@@ -359,6 +359,11 @@ public final class AetherEngine: ObservableObject {
     private var lifecycleObservers: [Any] = []
 
     public init() throws {
+        // Route FFmpeg's av_log output into EngineLog before any
+        // libav* entry point runs, so probe/load diagnostics land in
+        // Console.app + the host handler from the very first call.
+        FFmpegLogBridge.install()
+
         // Configure audio session for multichannel playback. AVPlayer
         // needs `.playback` + `.moviePlayback` + multichannel-content
         // enabled before its first asset opens, otherwise output is
--- a/Sources/AetherEngine/Diagnostics/EngineLog.swift
+++ b/Sources/AetherEngine/Diagnostics/EngineLog.swift
@@ -31,6 +31,14 @@ public enum EngineLog {
         /// category — session start/stop, lifecycle errors, anything
         /// that crosses subsystem boundaries.
         case engine
+        /// FFmpeg-internal diagnostics forwarded by `FFmpegLogBridge`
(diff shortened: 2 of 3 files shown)


superuser404notfound:
Merged as `c2ae3ef`, Sodalite's AetherEngine pin bumped in [`965abb62`](https://github.com/superuser404notfound/Sodalite/commit/965abb62). Thanks for this, the missing FFmpeg-internal log surface has been a real diagnostic gap for App Store builds and `aetherctl` runs alike. The defensive level re-check inside the callback (since `av_log_set_callback` bypasses the default gate) and the `AV_LOG_SKIP_REPEATED` flag in particular are the kind of detail that turns a "nice to have" into something we can actually leave on in shipped builds without drowning the log ring.

PR #40, opened 2026-06-16
feat(subtitles): opt-in advertise subtitle tracks as renditions for native pickers

l984-451:
## Summary

Adds an opt-in `LoadOptions.advertiseSubtitleRenditions` flag (default `false`). When set, AetherEngine advertises each of the source's subtitle tracks as a decoy (empty, cue-less) WebVTT rendition in the generated HLS master playlist, so AVKit's **native subtitle picker** lists them. The engine paints no cues for these renditions; the host renders the actual subtitle cues itself, consistent with the engine's host-renders subtitle design. This is purely a picker-population hook.

**Default off: with the flag unset, the master/media playlists are byte-identical to a build without this feature, and no subtitle lines are emitted. Existing behavior is unchanged.**

## Why

Hosts that paint their own subtitle cues (the engine decodes cues but does not render them) have no way to surface those tracks in AVKit's built-in Subtitles menu, because the tracks never appear in the HLS playlist AVPlayer sees. This adds a way to populate that native menu without changing how cues are rendered.

## What it does

- **`LoadOptions.advertiseSubtitleRenditions: Bool = false`** opt-in flag.
- **`AetherEngine.subtitleRenditions`** published mapping (`renditionID` / `name` / `language` / `trackIndex`). `renditionID` embeds the engine subtitle AVStream index so the host can correlate AVKit's selected legible option back to the `TrackInfo` whose cues it should decode and paint. Names are built as `"Language (FORMAT)"` with a codec-derived format label (SRT/ASS/VOBSUB/PGS/DVB/VTT/Text), preserving authored distinguishers (SDH/Forced/Commentary) and dropping bare echoes / remux auto-titles. Non-empty only while a load opted in; cleared otherwise and on stop.
- **`HLSVideoEngine`** serves a master playlist even for SDR titles while advertising (an `EXT-X-MEDIA:TYPE=SUBTITLES` line can only live in a master). This is deliberately **decoupled from `servingMasterPlaylist`**, the HDR-pipeline flag the host wires into HDR-only `AVPlayerItem` flags, which stays strictly the HDR decision. So an SDR title with subtitles serves a master (single SDR variant + subtitle group) but stays SDR for the host's HDR flags. **HDR/DV routing and SDR-without-subtitles are byte-for-byte unchanged.**
- **`HLSLocalServer`** emits one `#EXT-X-MEDIA:TYPE=SUBTITLES` line per rendition in a `subs` group (before the `STREAM-INF`, per Apple's convention) and adds `SUBTITLES="subs"` to the variant. It serves the decoy `/subs_<id>.m3u8` (a VOD WebVTT media playlist sized to the source duration) and `/subs_<id>.vtt` (an empty `WEBVTT` body). When `masterCodecs` is nil (SDR, no variant metadata) but renditions exist, `CODECS` is omitted from the `STREAM-INF` rather than skipping the variant, so the subtitle group still has a variant to attach to.

## Tests

Adds `MasterPlaylistSubtitleTests` covering:
- advertising emits the `EXT-X-MEDIA` subtitle lines, the `subs_<id>.m3u8` URIs, `SUBTITLES="subs"`, and orders media renditions before the `STREAM-INF`;
- SDR-with-subtitles serves a master and omits `CODECS`;
- the playlist is the bare `#EXTM3U\n` when the feature is off and there is no video metadata;
- a video master emits no subtitle lines when not advertising (gated on the opt-in).

`swift build` and `swift test` pass (full suite green). Branch is rebased on current `main`.

## Notes

This was developed and verified end to end against a tvOS host (text + PGS/VOBSUB bitmap tracks, language / on-off / style switching, same-language variants) using the host-paints-cues model. The change is additive and gated behind the new flag.

### DIFF
--- a/Sources/AetherEngine/AetherEngine+Loading.swift
+++ b/Sources/AetherEngine/AetherEngine+Loading.swift
@@ -233,7 +233,13 @@ extension AetherEngine {
             liveSourceCadenceHint: liveSourceCadenceHint,
             preopenedDemuxer: preopenedDemuxer,
             sourceReopenableByURL: !isCustomSource,
-            companionAudioReader: companionAudioReader
+            companionAudioReader: companionAudioReader,
+            // Decoy subtitle renditions for AVKit's native picker. Mapped
+            // from the published list the load probe populated (empty
+            // unless the active load set advertiseSubtitleRenditions).
+            subtitleRenditions: subtitleRenditions.map {
+                (renditionID: $0.renditionID, name: $0.name, language: $0.language)
+            }
         )
         session.onFirstHDR10PlusDetected = { [weak self] in
             Task { @MainActor in self?.handleHDR10PlusDetected() }
--- a/Sources/AetherEngine/AetherEngine+Probe.swift
+++ b/Sources/AetherEngine/AetherEngine+Probe.swift
@@ -64,6 +64,94 @@ extension AetherEngine {
         return makeSourceProbe(demuxer: demuxer, displayURL: displayURL)
     }
 
+    /// Build the decoy subtitle rendition list (native-picker path) from
+    /// the probed subtitle tracks. Each rendition's `renditionID` embeds
+    /// the engine subtitle AVStream index (`TrackInfo.id`) so the host can
+    /// correlate AVKit's selected legible option back to the track; the
+    /// `name` is a human label (the track's localized language name, then
+    /// its declared name, then "Subtitle N"); `language` is the ISO code
+    /// (fallback "und").
+    nonisolated static func makeSubtitleRenditions(
+        from tracks: [TrackInfo]
+    ) -> [SubtitleRendition] {
+        // The native picker lists each rendition by its NAME, and AVKit
+        // collapses entries sharing NAME + LANGUAGE. Build a consistent,
+        // self-describing label from the reliable signals (localized language +
+        // format derived from the codec), then append the container title only
+        // when it carries a real distinguisher (SDH / Forced / Commentary /
+        // Simplified), dropping bare echoes and remux auto-titles like
+        // "ENG (srt)". Number any labels that still collide so every track stays
+        // selectable. Result e.g. "English (SRT)", "English SDH (SRT)",
+        // "Chinese Simplified (SRT)", "English (VOBSUB)".
+        var renditions: [SubtitleRendition] = []
+        var usedNameCounts: [String: Int] = [:]
+        for (offset, track) in tracks.enumerated() {
+            let language = track.language?.trimmingCharacters(in: .whitespaces)
+            let langCode = (language?.isEmpty == false) ? language! : "und"
+            let localizedLang = (langCode != "und")
+                ? Locale.current.localizedString(forLanguageCode: langCode)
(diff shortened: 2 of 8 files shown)


DrHurt:
@l984-451 

Why not just add a custom subtitles button & menu to the AVPlayerViewController and bypass AVKit's HLS-driven track selection altogether? Same for audio tracks.

https://developer.apple.com/documentation/avkit/customizing-the-tvos-playback-experience

Much simpler that way.

Aether has fully custom subtitles and audio track selection pipelines and isn't meant to integrate with AVPlayer (view controller) the way you're suggesting.

l984-451:
@DrHurt
I like using Apple's built-in tools and methods whenever possible. A big draw to Aether is that it lets me use Apple's AVPlayerViewController with any media instead of using a custom view. Part of the fun for me in Rivulet is seeing what I can do to resemble an app that Apple would design.

DrHurt:
You’ll still be using AVPlayerViewController, just replacing the standard subtitles menu with a custom one. You can even use the same icon and the player UI will look identical to Apple TV+ 

Custom menus are ultimately an official feature of the native AVPlayerViewController whereas your suggestion is quite a hacky (albeit clever) way to force use a menu that’s not designed for this use case.

Ultimately up to Vincent if he wants to merge this into main. I’m just thinking out loud.

Good luck and looking forward to trying out rivulet once everything is implemented :)

superuser404notfound:
Thank you for this, it is genuinely well-built: opt-in, default-off byte-identical, the `servingMasterPlaylist` decoupling is reasoned correctly, and it ships with tests. I went through it carefully.

I am going to pass on merging it, and I want to explain why rather than just close it.

AetherEngine is deliberately "host renders subtitles": it decodes and publishes the cue stream (`subtitleCues`) and the track list (`subtitleTracks`), and selection/painting lives in the host. The native-looking picker you are after is already reachable from that surface without any engine change, via the route DrHurt pointed to: a custom subtitles menu on `AVPlayerViewController` ([Customizing the tvOS playback experience](https://developer.apple.com/documentation/avkit/customizing-the-tvos-playback-experience)) wired to the published `subtitleTracks`, painting `subtitleCues` on selection. That menu can use the same system icon and reads as a stock Apple TV+ player.

So this PR adds a second, heavier path (a public `LoadOptions` flag, a published rendition mapping, master-playlist + loopback-server machinery) to populate AVKit's built-in menu, where the host-side custom menu gets the same result with zero engine surface. For an engine whose subtitle design is host-rendered, I would rather keep that surface lean.

Two things so the work is not lost:

- If wiring the custom menu to the engine surfaces a genuine gap (something you need to correlate or drive the selection that the engine does not yet publish), open an issue for that and I will add a small, focused API. That is the kind of addition I want to carry.
- If down the line populating AVKit's stock menu specifically turns out to matter, this branch is the basis for it. One thing to fix first if so: the rendition `name`/`language` go unescaped into the quoted `#EXT-X-MEDIA` attributes, so a subtitle title containing `"`, CR, or LF would produce an invalid master playlist. It would need sanitizing.

Really appreciate the care here, and good luck with Rivulet.

PR #41, opened 2026-06-16
fix(hls): bound VOD cue-prewarm seek so a missing/out-of-bounds MKV C…

RomanLiberda:
## Summary

The VOD cue-prewarm seek can silently degrade into a multi-GB linear forward
scan (a de-facto hang) on MKV sources whose Cues index is missing or points
past EOF. This bounds the prewarm with a deadline so playback starts promptly,
falling back to the existing segment-plan path.

## What changed

- `HLSVideoEngine.cuePrewarmTimeout` (10s) now bounds the cue-prewarm seek.
- `AVIOReader`: optional read deadline (`beginReadDeadline` / `endReadDeadline`,
  `readDeadlineFired`). When armed, the read callback returns -1 once it passes,
  aborting the in-flight `avformat_seek_file` the same way `markClosed()` does
  for cancellation (loop-top check + deadline-bounded `winCond.wait`).
- `Demuxer.seekBounded(to:timeout:)`: arms the deadline around the seek and
  reports whether it was capped via the deadline flag — matroska can return
  success with a partial index after the abort, so the return code isn't
  authoritative.
- `HLSVideoEngine` uses `seekBounded` for the prewarm and falls through to its
  existing keyframe / uniform-stride plan from whatever keyframes were scanned.
- No public API changes (new symbols are internal and documented).

## Test plan

- Device / OS: macOS 26.2 (headless `aetherctl serve`) — verified; Apple TV 4K
  (gen 2) / tvOS 26 — original ~1h hang observed.
- Source media (container / video codec + profile / audio / HDR-DV):
  MKV / HEVC Main 10 (profile 2, level 5.1) / 4× DTS (5.1 EN, 5.1 EN, 2.0 RU,
  5.1 CS) + 2× PGS subs (EN, CS) / HDR10. 72.9 GB; SeekHead points Cues to
  ~77.4 GB — 4.5 GB past EOF.
- Result: time-to-`serving` drops from ~1 hour to ~11s; the loopback server
  serves and the segment producer streams normally. Healthy files are
  unaffected (deadline defaults to no-op; a sub-second Cues parse never trips it).

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Commit messages follow Conventional Commits (`fix(...)`)
- [x] Public API changes are intentional and documented (none — new symbols are internal)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ the public-API contract.
 ### Fixed
 
 - **Seek on the native loopback-HLS path no longer bounces back through the pre-seek position.** A seek wrote the target clock optimistically and flipped state back to `.playing` without waiting for AVPlayer's seek to physically land, so the 100 ms periodic time observer kept publishing the stale pre-seek clock until the (seconds-late) loopback seek completed — the reported time read the target, snapped back to the old position, then re-settled. `seek(to:)` now awaits the real AVPlayer completion, and the native host suppresses the periodic observer's stale reads while a seek is in flight, so the clock holds the target across the landing (AetherEngine#37).
+- **Hang on MKV sources with a missing or out-of-bounds Cues index.** When a file's Cues seek index is absent or points past EOF (truncated / mis-muxed remux), libavformat's matroska seek degrades the VOD cue-prewarm into a multi-GB linear forward scan — tens of minutes (a de-facto hang) on a large remote source, even though every byte range of the stream serves fine. The prewarm seek is now bounded by a deadline (`HLSVideoEngine.cuePrewarmTimeout`); on timeout it falls back to the existing keyframe / uniform-stride segment plan so playback starts promptly. Healthy files (Cues resolve in well under a second) are unaffected.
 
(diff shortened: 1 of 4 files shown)


superuser404notfound:
Merged, thank you. This is a clean, well-scoped fix and the writeup made it easy to review: bounding the prewarm seek with a deadline and treating `readDeadlineFired` (not `avformat_seek_file`'s return) as the authoritative "was it capped" signal is exactly right, since matroska keeps a partial keyframe index from the bytes it scanned before the abort. The strict no-op off the bounded path (`readDeadline = .distantFuture`) and the `min(connStallTimeout, readDeadline)` wait with the deadline-wake bail before the stall handling are the right details to get correct.

I rebased your branch on current `main` to resolve a `CHANGELOG.md [Unreleased]` collision with another fix that landed in the meantime (your commit is otherwise untouched, authorship preserved), and verified `swift build` + the full test suite (125 passing) on the rebased commit before merging. The 1h to 11s improvement on your 72 GB repro is a great catch.

Issue #77, opened 2026-06-26
Render embedded CEA-608/708 closed captions (eia_608) — currently no decode or passthrough path


ghost:
What should the engine do?
Render in-band CEA-608/708 closed captions (eia_608, carried in the H.264 SEI). Today such a track is exposed and selectable but produces no cues — the side-demuxer decode path can't open it and there's no native closed-caption passthrough — so a source whose only caption track is 608 shows nothing.

Motivating media or use case
Remote progressive MP4 streams: H.264 1080p + AAC + eia_608 (+ mjpeg cover). The 608 track is the only subtitles. With LoadOptions.preferredSubtitleLanguages: ["en"] (or a manual selectSubtitleTrack) the engine activates it, then can't decode it:

```
[Demuxer] stream[2] type=subtitle codec=eia_608 1920x1080
[AetherEngine] preferred-subtitle auto-select stream=2 langs=["en"]
[AetherEngine] embedded subtitle decoder open failed for stream=2
memprobe … subActive=true subCues=0 subTracks=1
```

EmbeddedSubtitleDecoder.init? returns nil — avcodec_find_decoder has no decoder for AV_CODEC_ID_EIA_608 (FFmpegBuild doesn't enable ccaption/cc_dec). The result is subActive=true / subCues=0: a track that's "on" but blank for the whole title. The same media shows captions in players that render in-band CC. (AetherEngine 4.5.1, macOS / tvOS, native loopback-HLS.)

Why the side-demuxer path can't serve this: CEA-608/708 data lives in the video stream's A53 SEI, not in independently-demuxable subtitle packets — a side demuxer opening only the subtitle stream has nothing to decode even if ccaption were built. The captions ride with the H.264 the engine already stream-copies into the loopback segments.

Suggested direction (and the one design caveat): since the SEI-bearing H.264 is already stream-copied into the segments, the lowest-cost route on the native path is to advertise an EXT-X-MEDIA:TYPE=CLOSED-CAPTIONS group (instead of the current hardcoded CLOSED-CAPTIONS=NONE) and let AVPlayer render the 608/708 itself — no decoder, no extra muxing, and it survives PiP/AirPlay/external display like the #55 mov_text path. This is purely additive: every other subtitle codec (SRT/ASS/VTT, PGS/DVB/DVD, mov_text) keeps using the side-demuxer / subtitleCues path unchanged; only in-band 608/708 uses the new route.

[...]

Issue #55, opened 2026-06-21
Native subtitle rendition (AVMediaSelection / WebVTT) so subtitles survive PiP, AirPlay & external display


edde746:
### What should the engine do?

Optionally emit the active subtitle as a native HLS subtitle rendition (`#EXT-X-MEDIA:TYPE=SUBTITLES` → WebVTT, served by `HLSLocalServer` from the already-decoded cues) and expose it through the legible `AVMediaSelectionGroup`, so AVPlayer renders the subtitle itself. That is the only way subtitles can survive Picture-in-Picture, AirPlay, and external-display playback, where a host-drawn overlay never appears.

### Motivating media or use case

**Scenario:** a tvOS/iOS client playing any title with an embedded or sidecar **text** subtitle (e.g. an MKV with an embedded ASS dialogue track, or a `.srt`/`.vtt` sidecar) where the user turns on **Picture-in-Picture**, **AirPlay**, or connects an **external display**.

**Today:** subtitles are surfaced to the host as decoded cues (`$subtitleCues` + the `preserveASSMarkup`/`assHeader` host-render contract from #30/#48) and the host paints them with swift-ass renderer. That works inline, but the overlay is a sibling layer **outside** the `AVPlayerItem`'s video output, so subtitles **disappear the moment playback moves to PiP / AirPlay / external display**.

**Why a host can't fix this itself** (so this has to live in the engine):
- A host overlay view (or `AVPlayerViewController.contentOverlayView`) is **not shown in PiP** by AVKit,
  and isn't sent over AirPlay / to an external display.
- The `AssVideoComposition` CoreImage `videoComposition` route can't be used either — **CIFilter video
  compositions are unsupported on HLS assets**, and the engine plays the loopback `media.m3u8` /
  `master.m3u8`. So there is no host-side path to get rendered subtitles into the frames AVPlayer emits.

The only thing the OS renders in PiP/AirPlay is what lives inside the `AVPlayerItem` — i.e. a native
legible selection. This also lines up with the `AVMediaSelection wiring is a tracked follow-up` note in
`AetherEngine.subtitleCues`'s doc comment; `VideoSegmentProvider.masterClosedCaptions` is currently
`"NONE"` and there's no `EXT-X-MEDIA:TYPE=SUBTITLES` rendition.

### Area

Public API surface

### Would you be willing to open a PR?

No

superuser404notfound:
[...]

PR #81, opened 2026-06-26
Render in-band CEA-608 closed captions (#77)

ghost:
## Summary

A source whose only caption track is an embedded **CEA-608** stream (`eia_608`, e.g. a QuickTime/MP4 `c608` track) currently can't render. FFmpegBuild ships no `ccaption` decoder, so the side-demuxer `EmbeddedSubtitleDecoder` open fails and the track sits **active-but-blank** (`subActive=true / subCues=0`) — selectable but producing no cues for the whole title.

This PR decodes those captions in-house and renders them through the existing host-overlay subtitle path, **without standing up a second demuxer, a second connection, or touching any existing pipeline**. The caption bytes are read off the segment producer's source connection that the engine already holds.

Closes the rendering half of #77. First cut: **608 field-1 / channel CC1**; CEA-708 (DTVCC) and field 2 are explicit follow-ons.

## What it does

- A track exposed as `eia_608` now renders its captions on the host overlay (`subtitleCues`), exactly like every other side-decoded subtitle codec (SRT/ASS/PGS/DVB).
- Works on local, remote (progressive MP4 over HTTP), and Jellyfin direct-stream sources.
- Captions appear **instantly** when the track is selected (no side-demuxer open/seek/prewarm), and survive **seek / reload / producer re-anchor (wedge)**.

## Architecture

Three pieces, all additive; the caption bytes ride the connection the producer already reads.

1. **`CCDataParser`** (new) — parses the bare `cc_data` triplet stream FFmpeg's MOV demuxer emits for a `c608` track (it strips the `cdat`/`cdt2` atom wrapper and hands downstream `(cc_valid|cc_type, b0, b1)` triplets). Pure byte parsing.
2. **`CEA608Decoder`** (new) — an in-house line-21 state machine: pop-on / roll-up / paint-on, PAC row addressing, mid-row codes, and the basic / special / extended West-European character sets. Odd-parity validation, doubled-control suppression, the character tables and PAC bounds are all validated against FFmpeg's `ccaption_dec.c` (the decoder mpv/VLC mirror) and covered by unit tests.
3. **`ClosedCaptionTap`** (new) — a read-only observer attached to the producer's source demuxer. The producer keeps the `eia_608` stream in its keep-set, hands each of its packets to the tap, then **drops it (never muxed)**. The tap decodes on the pump thread, owns the cue buffer, and publishes immutable snapshots to the MainActor, which mirrors them into `subtitleCues` *only while the CC track is the active subtitle*.

Because the tap owns the buffer and `makeProducer` re-threads the observer onto every producer (initial + restart), enabling CC mirrors the buffer instantly and seeks/reloads/wedges re-feed it automatically — no second connection, no per-seek re-arm. Snapshot delivery is sequence-guarded against out-of-order MainActor hops; the tap drops decoder + buffer state on an explicit seek (`requestReset()` from `seek()`).

## Minimal interference — existing pipelines intact

- **Byte-identical output.** The CC packet is *observed then dropped* in the pump's foreign-packet branch; it is never muxed. With no CC track, `closedCaptionStreamIndex == -1` and every site is a no-op — the loopback-HLS segments are bit-for-bit unchanged.
- **#55 native `mov_text` rendition untouched.** CC is *excluded* from the `textTracks` filter (a caption track can't become `tx3g`) and rendered on the overlay like the bitmap subtitle codecs. Consistent limitation: host-overlay only, so no PiP/AirPlay CC (same as PGS/DVB today).
- **All CC logic is externalised** in the three new files. Edits to existing files are small, additive, and observational:
  - `Demuxer.trackInfo` — fall back to `avcodec_get_name` when no decoder exists, so the track is identifiable (`eia_608`) instead of `"unknown"`. Cannot affect any track that has a decoder.
  - `AetherEngine+Probe` — add an `isEmbeddedClosedCaptionCodec` classifier beside `isBitmapSubtitleCodec`.
  - `AetherEngine+Loading` — exclude CC from the `mov_text` filter; arm the tap before `start()`.
  - `AetherEngine+Subtitles` / `AetherEngine` — route CC selection to mirror the tap's snapshot; reset the tap on seek.
  - `HLSVideoEngine` — thread the CC stream index + observer onto every producer (same pattern as the #55 `subtitleCueStores` threading).
  - `HLSSegmentProducer` — one keep-set entry + one read-only observer call.

## Scope / limitations (stated)

- 608 **field-1 / CC1** only; 708 (DTVCC packet reassembly) and field 2 are follow-ons.
- Host-overlay rendering only (no native `tx3g`, so no PiP/AirPlay CC) — same as the existing bitmap subtitle path.
- Handles the **demuxable caption track** carriage (`eia_608` / `c608`). Captions carried only in the video bitstream SEI (no separate track) are out of scope here.

## Extending this system (future work)

The architecture splits cleanly into **producer → parser → decoder → `subtitleCues` overlay**, and the overlay is the universal sink: anything that can emit timed text cues plugs in the same way. That makes several follow-ons cheap, in roughly increasing cost:

- **CEA-708 (DTVCC) and 608 field 2 / channels CC2–CC4 — same tap, no new I/O.** The tap already receives *all* `cc_data` triplets from the caption stream and currently filters to field-1 (`cc_type == 0`); `cc_type 2/3` (708) and `cc_type 1` (field 2) flow through the exact same observer. Adding them is purely a decode-side change: reassemble the DTVCC packets (708) or extend the existing 608 channel handling (field 2 / CC2–CC4) and publish to the same buffer. No second connection, no producer changes.
- **In-band text streams with no FFmpeg decoder — reuse the tap pattern.** Any demuxable caption/data stream the side-demuxer path can't open (a vendor caption codec, a teletext page, an ID3/`timed_id3` text feed) can keep its stream in the producer keep-set, hand its packets to a small purpose-built parser, and publish cues — identical shape to `ClosedCaptionTap`, byte-identical output preserved.
- **Burned-in / "baked-in" / telesync hardsubs — same sink, different source.** These live in the *video pixels*, not a data stream, so the packet-tap doesn't apply; they need a frame consumer + OCR (Vision on Apple platforms) running off the engine's existing software-decode / frame-extractor path, emitting recognized text as cues into `subtitleCues`. Heavier (it requires pixel decode), and best gated to on-demand, but it lands on the very same overlay — so the host UI, styling, positioning, and selection are unchanged.
- **Generated / external cue sources.** Translation, auto-captioning, or a fetched sidecar (SRT/VTT/ASS) are all just cue producers: parse/produce timed text off-thread and publish snapshots, exactly as the CC tap does.

In short: **`subtitleCues` is the extension point.** For in-band data, copy the read-only producer-tap; for pixel-baked text, add a frame-OCR producer; either way the rendering, host overlay, seek/reload survival, and instant-enable behavior come for free.

## Files

New: `Decoder/CCDataParser.swift`, `Decoder/CEA608Decoder.swift`, `AetherEngine+ClosedCaptions.swift`, plus `Tests/AetherEngineTests/CCDataParserTests.swift` and `CEA608DecoderTests.swift`.
Edited: `Demuxer/Demuxer.swift`, `AetherEngine+Probe.swift`, `AetherEngine+Loading.swift`, `AetherEngine+Subtitles.swift`, `AetherEngine.swift`, `Video/HLSVideoEngine.swift`, `Video/HLSSegmentProducer.swift`, `CHANGELOG.md`.

## Test plan

- **Unit:** 11 new tests — `CCDataParser` triplet parsing (cc_valid/cc_type, bounds) and `CEA608Decoder` (pop-on display/erase, roll-up on CR, doubled- and tripled-control suppression, odd-parity rejection, extended-glyph correctness, `reset()`). Full `AetherEngineTests` suite passes (221 tests, 0 failures) on macOS.
- **Manual (macOS):** H.264 1080p + AAC + `eia_608` MP4. Verified: captions render and stay in sync from cold start and on resume; survive short and long seeks; survive Jellyfin direct-stream reload; no A/V or audio regression. Selecting the track shows captions immediately.

## Acknowledgements

Externalise-subtitles steer by DrHurt.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ the public-API contract.
 
 ## [Unreleased]
 
+### Added
+
+- **In-band CEA-608 closed captions from a demuxable caption track, rendered through the host overlay (#77).** A source whose only caption track is an embedded CEA-608 stream (`eia_608`, e.g. a QuickTime/MP4 `c608` track) previously could not render: FFmpegBuild ships no `ccaption` decoder, so the side-demuxer `EmbeddedSubtitleDecoder` open failed and the track sat active-but-blank (`subActive=true / subCues=0`). The engine now reads that caption track's `cc_data` off the segment producer's existing source connection — a read-only observer keeps the `eia_608` stream in the demuxer's keep-set, hands each of its packets to an external `ClosedCaptionTap`, then drops it (never muxed → the loopback-HLS segment output is byte-identical to the no-CC path). An in-house CEA-608 decoder (pop-on / roll-up / paint-on, PAC row addressing, mid-row codes, and the basic / special / extended West-European character sets; odd-parity validation, doubled-control suppression and the character / PAC tables validated against FFmpeg's `ccaption_dec.c`) turns the bytes into cues published on the same `subtitleCues` host-overlay path as every other side-decoded subtitle codec. Because the tap owns the cue buffer and rides the producer (re-threaded onto every restart via `makeProducer`), captions appear instantly on enable — no second demuxer, no extra connection — and survive seek / reload / wedge. The native `mov_text` rendition (#55) is untouched: CC is excluded from that path (it can't become `tx3g`) and rendered through the overlay like the bitmap subtitle codecs, so — as with those — it is host-overlay only (no PiP / AirPlay CC). First cut: 608 field-1 / channel CC1; CEA-708 (DTVCC) and field 2 are follow-ons. Thanks to DrHurt for the externalise-subtitles steer.
+
 ## [4.5.4] — 2026-06-26
 
 ### Fixed
--- a/Sources/AetherEngine/AetherEngine+ClosedCaptions.swift
+++ b/Sources/AetherEngine/AetherEngine+ClosedCaptions.swift
@@ -1,0 +1,181 @@
+import Foundation
+import AVFoundation
+import Libavformat
+import Libavcodec
+import Libavutil
+
+/// In-band CEA-608 closed-caption support for a demuxable caption track (`eia_608` / QuickTime `c608`) (#77).
+///
+/// Externalised tap: a read-only observer attached to the `HLSSegmentProducer`'s source demuxer (the
+/// connection the engine already reads reliably). The producer keeps the `eia_608` caption stream, hands
+/// each of its packets to the tap, then drops it (never muxed → output byte-identical). The tap decodes
+/// on the pump thread
+/// and **owns the cue buffer**, publishing an immutable snapshot to the MainActor whenever it changes.
+/// The engine mirrors that snapshot into `subtitleCues` while the CC track is the active subtitle.
+///
+/// Because the tap holds the full buffer, enabling CC mid-playback shows captions instantly (mirror the
+/// snapshot — no producer restart), and a seek that re-pumps an overlapping region can't create duplicate
+/// cues (the single buffer owns them). `makeProducer` re-threads the observer onto every restart, so it
+/// survives seek/reload/wedge with no second connection.
+///
+/// First cut: 608 field-1 / CC1 only (see `CEA608Decoder`). Thread-confined to the pump thread for decode;
+/// `@unchecked Sendable` for the observer-closure boundary. Snapshot publication hops to the MainActor.
+final class ClosedCaptionTap: @unchecked Sendable {
+    private weak var engine: AetherEngine?
+    let ccStreamIndex: Int32
+
+    private let decoder = CEA608Decoder()
+    private var lastPTS: Double = -1
+
+    // Pump-thread-owned cue buffer.
+    private var cues: [SubtitleCue] = []
+    private var openCueID: Int?
+    private var nextID = 0
+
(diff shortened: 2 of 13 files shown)


ghost:
https://github.com/user-attachments/assets/1657053b-21e9-4b48-9c4b-e419b04596c5

superuser404notfound:
Merged and shipped in [4.6.0](https://github.com/superuser404notfound/AetherEngine/releases/tag/4.6.0). Thanks @reckloon, this is a clean, well-tested implementation that lands exactly the externalised approach from the discussion: the read-only producer tap keeps the loopback segment output byte-identical, and the in-house CEA-608 decoder matches FFmpeg's `ccaption_dec.c`. I verified the parity handling, the doubled / tripled-control suppression (this FFmpeg revision does not self-clear `prev_cmd` in the dedup branch, so the decoder matches it exactly), and the full basic / special / extended character and PAC tables against the reference locally. Build is clean and all 226 tests pass on the merge.

Follow-ons noted for later: CEA-708 (DTVCC), 608 field 2 / channels CC2 to CC4, and the SEI-only carriage. Thanks also @DrHurt for the externalise-subtitles steer.

PR #139, opened 2026-07-19
fix(audio): set correct session output channels based on source. 

### LINKED ISSUE #138: Preferred audio output channels is not set based on source audio channels
### What happened?

When `AVAudioSession` is established, the preferred output channels is set to max channels. It's not based on the source audio channels. This results in stereo audio tracks being played on a 7.1 AVR as 7.1 channels (which does not allow Dolby Pro Logic / DTS Neural X up mixers to process the audio).

### Steps to reproduce

1. Play a video on the SoftwarePlaybackHost path that has stereo audio (live TV channels etc)
2. Confirm connected AppleTV AVR reports max channels output (not stereo audio)

5.8.8

### Host app

Sodalite

### Platform

tvOS

### OS version

tvOS 26.5

### Device / chip

Apple TV 4K (3rd Gen)

### Playback path

Software (dav1d) host

Live AU TV channels / interlaced VOD with stereo tracks.

### Error codes / log lines

```text

```

### Anything else

_No response_

nathanpiper:
## Summary

Sets `AVAudioSession.setPreferredOutputNumberOfChannels` based on source active audio track's actual channels, defaulting to max channels when it can't be determined.

Closes #138

## What changed

- activateRendererAudioSession preferred output channels logic.

## Test plan

- Device / OS: Apple TV 4K (3rd Gen)
- Source media (container / video codec + profile / audio / HDR-DV): recorded mpegts, live channels and AV1 multi-track movie stereo/5.1  (to force software path decoding)
- Result: Initiating video and switching audio (in Sodalite) sets output channels correctly

## Checklist

- [x] `CHANGELOG.md` updated
- [ ] Public API changes are intentional and documented

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ the public-API contract.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Output audio channels set based on source (SoftwarePlaybackHost).** `AVAudioSession.setPreferredOutputNumberOfChannels` is now set based on active audio track's actual channels, defaulting to max channels when it can't be determined. 
+
 ## [5.8.8] - 2026-07-19
 
 ([release notes](https://github.com/superuser404notfound/AetherEngine/releases/tag/5.8.8))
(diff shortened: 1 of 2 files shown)


superuser404notfound:
Merged and shipped in 5.8.9. Thanks for the clean fix and the on-device verification, Nathan.

The renderer paths (SoftwarePlaybackHost, AudioPlaybackHost) now request a preferred output channel count that matches the active audio track instead of pinning to the route maximum, so a stereo source reaches the AVR as stereo and its Dolby Pro Logic II / DTS Neural:X upmixer engages again, while genuine multichannel sources still get every channel. Both consumers (Sodalite, AetherPlayer) are already bumped to 5.8.9.

Issue #127, opened 2026-07-16
iOS: quick app switches pay a full pipeline rebuild — grace period before paused-background teardown, warm reload, deferred pre-ready seeks, public readiness signal


dlev02:
### Context

AetherEngine on **iPhone**. The engine's background lifecycle was (I'm assuming, could be wrong) shaped by tvOS since that's the first platform for Sodalite (where backgrounding mid-playback is rare and the wedge-safety teardown is exactly right), but on iPhone or iPad the single most common flow is: **pause → swipe to Home → answer a text or do something real quick → come back 10–30 seconds later**. Today that quick switch pays a full pipeline rebuild, which on a slow single-connection debrid source takes 10–30 s to get back to a ready, correctly-positioned session.

### What happens today (device logs, 5.0.3, HEVC MP4 over a single-connection debrid CDN, mediocre Wi-Fi)

1. `didEnterBackground` while paused → `teardownVideoForBackground()` releases the pipeline immediately (correct per the wedge-safety contract).
2. On return the host calls `reloadAtCurrentPosition()`, which re-pays everything:
   - demuxer reopen: ~3–6 s to first byte on this source (plus a reconnect)
   - segment plan scan (145 IRAPs) + cue prewarm seek
   - AVPlayer item re-load → `readyToPlay` took 9–30 s across runs
3. Positioning is **correct** (`initial producer anchored at idx=1 (startPosition=13.01s)`, host load `startPos=13.01s`) — the cost is purely the rebuild latency. For a 15-second app switch, a 30-second resume is a rough exchange rate, and it happens on every switch.

### Proposals (roughly in order of impact)

**1. Grace period before the paused-in-background teardown (iOS).**
Hold the pipeline under a `UIBackgroundTask` assertion for a short window (~15 s) before running `teardownVideoForBackground()`; cancel on `didBecomeActive`. Wedge-safety is preserved: the assertion keeps the app genuinely running for the whole window, and the teardown runs at expiry (or in the assertion's expiration handler) — the pipeline is never alive across an idle suspension. We prototyped this locally against 5.0.3 (schedule in the `.teardownVideo` case + the pause-while-backgrounded path, cancel in `didBecomeActive`, re-evaluate `backgroundAction` at expiry since PiP/lock-screen-play can change mid-window) and quick app switches become seamless — the host reload never runs because `playbackBackend` never drops to `.none`. Happy to send the patch as a PR if the direction sounds right.

**2. Warm reload across teardown.**
[...]

PR #179, opened 2026-07-21
feat(demux): surface container (Matroska/MP4) chapters as mediaChapters

natedogg058:
Reads AVFormatContext.chapters in the probe demuxer and publishes them as
engine.mediaChapters, giving hosts a chapter list for ordinary file/URL
sources to build chapter navigation UI on - previously only disc
(DVD/Blu-ray) sources exposed chapters via discChapters.

- Demuxer.mediaChapterInfos(): sorted by start, re-ided sequentially,
  "Chapter N" fallback naming (matching the disc naming). Declared end
  times are trusted only for the last entry - real-world MKV muxes
  routinely write end == start - earlier durations run to the next start.
- Suppressed for disc sources: those keep their playlist/IFO chapters on
  discChapters with title-relative seek semantics; both lists never
  populate at once, so hosts can bind one picker.
- Wiped in both engine reset paths; preserved across the audio-switch /
  background reload seam via a pre-stopInternal snapshot (the source URL
  is unchanged, so re-probing would be waste), while disc chapters keep
  their existing reopened-demuxer recapture (a title switch legitimately
  changes them).

startSeconds values are content timestamps a host passes straight to
seek(to:). Shipped in a production tvOS consumer driving a chapter rail
on both this engine and a libmpv twin.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

---
Validated: builds for generic tvOS Simulator (arm64) against this repo's own manifest; running in a production tvOS consumer (custom chapter-rail UI over this API alongside a libmpv twin reading `chapter-list`, so the surface was designed engine-agnostic).

### DIFF
--- a/Sources/AetherEngine/AetherEngine+Loading.swift
+++ b/Sources/AetherEngine/AetherEngine+Loading.swift
@@ -1310,6 +1310,10 @@ extension AetherEngine {
         let wasOnSoftwarePath = (playbackBackend == .software)
         // Preserve codec so the decoder label can be reconstructed without re-probing.
         let preservedVideoCodec = lastDetectedVideoCodec
+        // Container chapters belong to the (unchanged) source URL, so an audio-switch / background
+        // reload republishes the load-time snapshot rather than re-probing. Disc chapters are NOT
+        // snapshot here: a title switch changes them, so they recapture from the reopened demuxer below.
+        let preservedMediaChapters = mediaChapters
         let reloadStart = DispatchTime.now()
         EngineLog.emit("[AetherEngine] reload: stopInternal start", category: .engine)
         // resetDisplayCriteria: false: video format is unchanged; resetting triggers a full waitForSwitch Stage 2 timeout (5 s at the 2026-05-26 device test, ~2 s cap since #117; Bose SLIII A2DP + 4K HDR10 PQ: each switch added ~12 s black-screen). On the same route a panel SDR drop during the reset window failed the PQ variant with AVFoundationErrorDomain -11868 / CoreMediaErrorDomain -17223.
@@ -1510,6 +1514,8 @@ extension AetherEngine {
             }
         } else {
             activeDiscTitleID = nil
+            // Non-disc reload: restore the container chapters stopInternal wiped (same source, same chapters).
+            mediaChapters = preservedMediaChapters
         }
 
         // Re-arm subtitle: sidecar branch wins because loadedSidecarURL is set only for sidecar sources.
--- a/Sources/AetherEngine/AetherEngine.swift
+++ b/Sources/AetherEngine/AetherEngine.swift
@@ -192,6 +192,11 @@ public final class AetherEngine: ObservableObject {
     /// Chapters of the selected title. Empty until Blu-ray chapter parsing ships (Phase 2); declared
     /// now so hosts can bind the picker against a stable API.
     @Published public internal(set) var discChapters: [ChapterInfo] = []
+    /// Container (Matroska/MP4) chapters of the loaded source, from the probe demuxer at load. Empty
+    /// when the container declares none and for disc sources (whose chapters publish on `discChapters`
+    /// with title-relative seek semantics). `startSeconds` values are content timestamps a host passes
+    /// straight to `seek(to:)`.
+    @Published public internal(set) var mediaChapters: [ChapterInfo] = []
     /// The id of the title the disc demuxer should (re)open with. Mirrors `selectedDiscTitle?.id` but
     /// kept as plain state so it survives the stopInternal inside a reload and threads into audio-switch /
     /// background-resume reopens (a URL-disc reopen with no id would silently revert to the main title).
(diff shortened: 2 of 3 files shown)


superuser404notfound (changes requested):
Thanks for this, and for the careful writeup. The quality here is well above the usual first PR: the disc suppression gate, the end-times-are-lies handling for real-world MKV muxes, and the naming parity with the disc chapters all match how this codebase thinks. I verified the core semantic claim before reviewing the mechanics: for non-disc sources `sourcePresentationOrigin` is 0, so `seek(to:)` and `currentTime` run on the container source-PTS axis on both backends, which is exactly the axis `AVChapter` starts live on (including sources with a non-zero `start_time`). "Pass straight to `seek(to:)`" holds generally, not just for the common start-at-zero case. I also ran the full suite on this branch merged onto current main (933/933 green) and approved the CI run, which is green as well.

One structural change request, then two smaller asks.

**The stopInternal wipe + snapshot loses chapters across the iOS background seam.** Your snapshot in `reloadWithAudioOverride` covers the audio-switch reload, where stopInternal runs inside the helper. But on iOS, `teardownVideoForBackground` (and the #127 expiration backstop `expireBackgroundGraceNow`) call `stopInternal` directly, outside the reload helper, and the designed resume path is `reloadAtCurrentPosition()` on foreground return. By the time that reload takes its snapshot, `mediaChapters` is already empty, so the restore restores nothing and the chapter list is gone after every background round trip. Your tvOS consumer can't see this because that teardown path is iOS-only.

The engine already has a pattern for exactly this situation, spelled out in the `stop()` comment for `fontAttachments`: session-scoped source-descriptive state "must survive stopInternal", because reload seams skip the probe. `audioTracks`, `subtitleTracks`, `metadata`, and `duration` all follow it; the disc lists are the deliberate exception because a title switch legitimately changes them, which does not apply to container chapters of an unchanged URL. So:

- drop `mediaChapters = []` from `stopInternal`
- add `mediaChapters = []` to the clear block in `stop()` instead (next to `fontAttachments`)
- keep the wipe in the `load()` prep block as you have it
- delete both `+Loading.swift` hunks (snapshot and restore), they become unnecessary

Net effect: smaller diff, the iOS seam fixed for free, and the chapter rail stays published during background/PiP the same way the track lists do.

**Tests.** The mapping logic (NOPTS/timebase guards, sort, sequential re-id, trust-end-only-for-last, title fallback) is pure and deserves fixture-free coverage. The precedent is `Demuxer.resolveBitrate` plus `DemuxerBitrateTests`: factor the raw-entries-to-`ChapterInfo` mapping into a static helper on `Demuxer` and test it directly. While you are in there, `tb.num >= 0` should be `tb.num > 0`: a zero numerator currently collapses the entry to start 0 instead of skipping it.

**Docs.** `mediaChapters` should join the published-property list in the README (next to `$discChapters`) and get a sentence in the chapters section of `docs/formats.md`, including the axis contrast with `selectChapter`. The `ChapterInfo` doc comment ("offset from the title's start") is now too narrow with a second producer of the type; worth a line covering both semantics. No CHANGELOG action needed, that is handled at release time.

One optional thought, no action required: with end == start on the last entry the final chapter gets `durationSeconds` 0. If you would rather not fall back to the container duration there, documenting 0 as "runs to end of file" on the property would also do.

Happy to merge once the stopInternal restructure and tests are in.

superuser404notfound:
Merged, thanks for this. Chapters for ordinary containers were a real gap, the read side is right where it belongs, and suppressing them for disc sources so exactly one of the two publishers is ever populated is the detail that makes the API honest rather than ambiguous.

I merged it and then pushed the review points myself rather than sending it back, since the fix turned out to shrink the diff:

**The wipe moved from `stopInternal` to `stop()`.** The snapshot/restore pair around the reload helper covered the audio-switch path, but `teardownVideoForBackground` and the #127 expiry call `stopInternal` outside that helper, so an iOS background roundtrip emptied the list for the rest of the session and the foreground reload then snapshotted the already-emptied list. Container chapters belong to the source URL rather than to the pipeline, which is the same rule `fontAttachments` follows, so both `+Loading` hunks came back out. tvOS was never affected, which is why the field validation passed cleanly.

**`tb.num >= 0` became `> 0`.** A zero numerator scales every timestamp to zero and collapses the whole list onto 0 s.

**The last chapter falls back to the container duration** when its declared end is degenerate. That is the same mux bug the next-start rule already handles, landing on the one entry the rule cannot cover, and it showed up as a zero-length chapter in a host UI.

**The mapping is now a pure function** over the values read off `AVChapter`, following the `Demuxer.resolveBitrate` precedent, with `ContainerChapterMappingTests` covering ordering, ids, naming and all three duration rules. `ChapterInfo` documents both time axes now, and README / `docs/formats.md` describe the publisher.

One thing I deliberately left alone: `selectChapter(id:)` stays disc-only and no-ops for a container chapter id. Extending it means adding the right content-start base, and whether a non-disc source needs one differs by backend, so that wants a device test rather than a same-commit convenience. It is documented as disc-only for now.

Shipping in the next release.

PR #180, opened 2026-07-21
feat(native): own an MPNowPlayingSession for the native video path

natedogg058:
The audio AVPlayer path already owns an auto-publishing MPNowPlayingSession
(AudioAVPlayerHost) because the shared MPRemoteCommandCenter /
MPNowPlayingInfoCenter aren't reliably bound to a bare AVPlayer: a
background pause drops the app as active Now-Playing and the shared center
stops receiving commands. The native VIDEO path had the same exposure with
no equivalent - custom-UI hosts (no AVPlayerViewController) got the
half-working bare-AVPlayer registration.

NativeAVPlayerHost now owns the same sanctioned session pattern:

- MPNowPlayingSession bound to the host's AVPlayer, auto-publish ON, so
  elapsed/rate/duration come from player truth and nobody writes the
  shared info center.
- The session persists with the host across native->native reloads
  (issue #15), so MediaRemote registration survives the reload seam that
  previously blanked the system card.
- Per-item identity metadata staged via setNowPlayingInfo and replayed
  onto every fresh AVPlayerItem, so readiness-gate master reloads, media
  fallbacks, and in-place swaps keep the card.

Engine surface mirrors the audio contract exactly:
videoNowPlayingSession (hosts register transport commands on its
remoteCommandCenter) and setVideoNowPlayingInfo (caller-managed, empty
clears, safe before load(), replayed at host creation).

Shipped in a production tvOS consumer with a custom player UI; commands
and card verified against pause/resume, seek, and gate-reload seams.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

---
Validated: builds for generic tvOS Simulator (arm64) against this repo's own manifest; running in a production tvOS consumer whose host registers transport commands on `videoNowPlayingSession.remoteCommandCenter` and stamps identity via `setVideoNowPlayingInfo`.

### DIFF
--- a/Sources/AetherEngine/AetherEngine+Loading.swift
+++ b/Sources/AetherEngine/AetherEngine+Loading.swift
@@ -145,6 +145,9 @@ extension AetherEngine {
         if !pendingExternalMetadata.isEmpty {
             host.setExternalMetadata(pendingExternalMetadata)
         }
+        if !pendingVideoNowPlayingInfo.isEmpty {
+            host.setNowPlayingInfo(pendingVideoNowPlayingInfo)
+        }
         self.nativeHost = host
         // A surface bound BEFORE load ran presentCurrentLayer() while nativeHost was still nil
         // (no-op); without this re-present nothing ever attaches host.playerLayer and AVPlayer
@@ -727,6 +730,9 @@ extension AetherEngine {
         // Forward pre-load externalMetadata so the AVPlayerItem picks it up before AVPlayer assigns it.
         if !pendingExternalMetadata.isEmpty {
             host.setExternalMetadata(pendingExternalMetadata)
+        }
+        if !pendingVideoNowPlayingInfo.isEmpty {
+            host.setNowPlayingInfo(pendingVideoNowPlayingInfo)
         }
         self.nativeHost = host
         applyDesiredVolume(to: host)
--- a/Sources/AetherEngine/AetherEngine.swift
+++ b/Sources/AetherEngine/AetherEngine.swift
@@ -2914,6 +2914,30 @@ public final class AetherEngine: ObservableObject {
         audioAVPlayerHost?.setExternalMetadata(items)
     }
 
+    #if os(tvOS) || os(iOS)
+    /// MPNowPlayingSession for the native VIDEO path, or nil (software path / no host). Same contract
+    /// as `audioNowPlayingSession`: the host app registers transport commands on
+    /// `remoteCommandCenter` and stages identity metadata via `setVideoNowPlayingInfo`; the session
+    /// auto-publishes elapsed/rate/duration from the AVPlayer and survives native->native reloads
+    /// with the host (issue #15), so system Now-Playing ownership holds across a background pause.
+    public var videoNowPlayingSession: MPNowPlayingSession? {
+        nativeHost?.nowPlayingSession
+    }
+    #endif
+
+    /// Staged per-item Now-Playing dictionary for the native video path. Replayed at host creation
+    /// and onto every fresh AVPlayerItem (gate reloads, media fallback, in-place swaps). Caller-managed
+    /// like the audio variant: pass an empty dict to clear.
+    var pendingVideoNowPlayingInfo: [String: Any] = [:]
+
+    /// Stage the system Now-Playing identity dictionary for the native video path (MPMediaItemProperty
+    /// keys + a force-decoded, @Sendable-wrapped MPMediaItemArtwork). Elapsed/rate/duration keys are
+    /// unnecessary — the session merges the player truth. Safe before load(); replayed at host creation.
+    public func setVideoNowPlayingInfo(_ info: [String: Any]) {
+        pendingVideoNowPlayingInfo = info
+        nativeHost?.setNowPlayingInfo(info)
+    }
+
     #if os(iOS) || os(tvOS)
     /// Staged per-item Now-Playing dictionary for the audio AVPlayer path. Replayed at host creation.
     var pendingAudioNowPlayingInfo: [String: Any] = [:]
--- a/Sources/AetherEngine/Native/NativeAVPlayerHost.swift
+++ b/Sources/AetherEngine/Native/NativeAVPlayerHost.swift
@@ -2,6 +2,9 @@ import Foundation
 import AVFoundation
 import AVKit
 import Combine
+#if os(tvOS) || os(iOS)
+import MediaPlayer
+#endif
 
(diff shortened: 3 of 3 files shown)


superuser404notfound (changes requested):
Thanks for another well-researched contribution, and for field-validating it in a shipping tvOS consumer. The direction is right: this is exactly the engine-side version of what `docs/formats.md` (Now-Playing note) already tells custom-transport hosts to do by hand, the staged dict correctly survives `stopInternal` (same reasoning as the `fontAttachments` precedent we discussed on #179), and both `NativeAVPlayerHost` creation sites in `+Loading` got the replay. I merged the branch with current main locally and the full suite passes (933/933).

One architectural change is required before this can land, plus a few smaller items.

## 1. The session must be opt-in, default off (blocking)

`AudioAVPlayerHost` can own its session unconditionally because the audio path is always a bare `AVPlayer` with a custom UI. The native video path is not: it is also consumed by `AVPlayerViewController` hosts, and AVKit owns its own Now-Playing registration there (via MediaRemote internally, not `MPNowPlayingSession`). The audio host's own header comment quotes the WWDC22 guidance this PR now violates for those consumers: "don't use MPNowPlayingSession on tvOS WHEN USING AVKit (AVKit owns its own)".

As written, every `NativeAVPlayerHost` init creates a session bound to the real player, turns auto-publish on, and calls `becomeActiveIfPossible`, for every consumer, including AVKit ones that never stage identity metadata. That gives an AVKit host two Now-Playing publishers on the same player: an active, auto-publishing session with empty identity competing with AVKit's registration. Plausible failure modes are a system card that loses title/artwork, or transport commands routing to the session's command center where nobody registered handlers, deadening Control Center controls that work today. There is also a second collision: `docs/formats.md` currently instructs custom-transport hosts to own their own `MPNowPlayingSession` against `currentAVPlayer`; any host that followed that advice would now have two sessions on one player as well.

Requested shape: an engine-level flag (for example `public var ownsVideoNowPlayingSession: Bool = false`, set before `load()`), read at host creation. `NativeAVPlayerHost` takes it in init, `nowPlayingSession` becomes `MPNowPlayingSession?`, created only when opted in, and `videoNowPlayingSession` stays nil otherwise. Default off keeps every existing consumer byte-identical; your app opts in with one line. An engine property (rather than a `LoadOptions` field) matches the session's lifetime, which spans native-to-native reloads.

## 2. Clear the staged dict in public `stop()`

`stop()`'s cleanup block clears `pendingExternalMetadata` so a stopped session cannot leak the previous title's metadata into the next load. `pendingVideoNowPlayingInfo` needs the same line there. Without it, `stop()` followed by `load()` of a different title replays the previous title's identity (title/artwork) onto the new item until the caller re-stages. Keep it surviving `stopInternal` as you have it, that part is correct.

## 3. Platform guards, parity with the audio API

`pendingAudioNowPlayingInfo` / `setAudioNowPlayingInfo` are wrapped in `#if os(iOS) || os(tvOS)`; the new `pendingVideoNowPlayingInfo` / `setVideoNowPlayingInfo` are not. The package builds on macOS (tests, aetherctl, a macOS consumer), and `docs/architecture.md` documents macOS as shared-centers territory, so macOS should not gain a silently no-op public API. Please mirror the audio guards.

## 4. Re-assert active on load (parity with audio)

The audio path calls `becomeActiveNowPlaying()` at track start to reclaim ownership if it was lost, on top of the init-time activation. The video session only activates at host creation, and the reuse path (native-to-native reload, and foreground return after the iOS background teardown, which preserves the host shell precisely for Now-Playing continuity) never re-asserts. Please mirror the audio re-assert on the native video load path, gated on the opt-in.

## 5. Docs

- README "Info panel / Now Playing" block: add `setVideoNowPlayingInfo` and `videoNowPlayingSession` next to `setExternalMetadata`.
- `docs/architecture.md`: a video-path paragraph next to the existing audio-session one (including the opt-in and the AVKit caveat).
- `docs/formats.md` Now-Playing note: point custom-transport hosts at `videoNowPlayingSession` instead of telling them to roll their own session against `currentAVPlayer`.
- Doc comment on `videoNowPlayingSession`: unlike the persistent audio host, the video host (and therefore the session instance) is recreated after `stop()`, so hosts must re-fetch the session and re-register commands after each `load()`, not once per app run.

## 6. Tests

The session itself is compiled out on macOS CI, but the decision logic the rework introduces is testable fixture-free (the #179 pattern): opt-in off yields a nil session, staging before `load()` is replayed at host creation, and an empty dict clears. Whatever subset of that is practical headless is enough.

CI note: I approved the workflow run for this branch, so you will see platform builds on your next push.

G00380316:
So funny how I just figured this out now too literally watching a show with AirPods but can't play from it 😅 adding my own MediaPlayer Session to my app

superuser404notfound:
Merged, thanks. The `AudioAVPlayerHost` shape was the right one to mirror, and staging the identity so it replays onto every fresh item is exactly what the readiness-gate master reload and the in-place PiP swap need; both would otherwise drop the card mid-session.

One change was required before this could ship, and I made it rather than sending the PR back.

**The session had to become opt-in.** It was created and activated in every `NativeAVPlayerHost`, but that host is consumed by `AVPlayerViewController` hosts as well as custom-transport ones, and AVKit owns Now-Playing there through its own private MediaRemote registration. The WWDC22 line quoted in the `AudioAVPlayerHost` comment is specifically "don't use MPNowPlayingSession on tvOS **when using AVKit**": the audio host can ignore it because it is always a bare AVPlayer, the video host cannot. Under AVKit an unconditional session costs that host its identity card, its `externalMetadata` and its working remote commands, since the commands route into a session with no handlers registered on it.

So ownership is now the host's call: `ownsVideoNowPlayingSession` (default false) is read when a native host is created and threaded into `NativeAVPlayerHost`, whose session is optional. The identity write and the per-item replay are gated on it too, so an AVKit host cannot end up feeding a second, host-authored identity into AVKit's own publication. Nothing changes for a custom-UI host like yours beyond one line before `load()`.

Three smaller things came with it:

- `becomeActiveNowPlaying()` re-asserts on the load path, matching the audio path. A host preserved across a native->native reload never re-runs `init`, so a session another app took over in the meantime was never claimed back.
- `pendingVideoNowPlayingInfo` clears in `stop()` next to `pendingExternalMetadata`. Surviving `stopInternal` stays correct (a reload keeps the card through the seam), but leaving playback should not carry the old title into the next session.
- The public API is behind `#if os(tvOS) || os(iOS)` like its audio counterpart, instead of compiling on macOS as a silent no-op.

`docs/formats.md` used to tell custom-transport hosts to hold their own session against `currentAVPlayer`; it now points at this API instead, which is the better answer and the reason the advice existed in the first place.

Shipping in the next release.

Issue #235, opened 2026-07-28
Stylized ASS subs are not rendered properly


fivepandasna:
### What happened?

Heavily styled ASS subtitles do not render correctly with libass (applies to Sodalite). Some subtitle events are missing, causing parts of the subtitle track to disappear or render inconsistently.

### Steps to reproduce

1. Play a video with heavily styled ASS subtitles (I can provide a sample track if needed)
2. Verify that some subtitle styles, layers, or dialogue events are missing or rendered incorrectly

### AetherEngine version or commit SHA

5.20.7

### Host app

Sodalite

### Platform

iOS

### OS version

iPadOS 26.5.2, iOS 26.5.2

### Device / chip

iPhone 17, iPad mini gen 7

### Playback path

Native AVPlayer

### Source media (for playback bugs)

*No response*

### Error codes / log lines

```text
```

### Anything else

I added logging to a custom player using AetherEngine and compared the processed events against the source ASS file. By the time playback had reached ~30.9s, `processEvent` had only been called 245 times, even though the ASS file contains 2,054 `Dialogue` events with start times less than or equal to the latest timestamp received by the player (~90.4s).

The number of processed events closely matches the number of unique timestamp ranges: there are only 268 unique `(start, end)` timestamp pairs in that same range. This suggests that multiple valid ASS dialogue entries sharing identical timestamps are being collapsed somewhere before `AetherPlayerWrapper` receives them.

While investigating, I found the following in `SubtitlePacketStore.appendLocked`:

```swift
let insertAt = entries.firstIndex { $0.ptsSeconds >= ptsSeconds } ?? entries.count
if insertAt < entries.count, entries[insertAt].ptsSeconds == ptsSeconds {
    bytes -= entries[insertAt].payload.count
    entries[insertAt] = entry        // overwrites existing entry
} else {
    entries.insert(entry, at: insertAt)
}
```

This appears to make `ptsSeconds` effectively unique, meaning subtitle packets with identical PTS values overwrite each other instead of being stored as separate entries.

superuser404notfound:
Fixed in [5.25.2](https://github.com/superuser404notfound/AetherEngine/releases/tag/5.25.2), via #237.

[...]

PR #237, opened 2026-07-28
fix(subtitles): don't collapse distinct packets that share a PTS

fivepandasna:
## Summary

Refs #235

## What changed

- Only collapse subtitle packets when both PTS and payload bytes match.
- Now only real duplicated are collpased, while heavy ASS stylization is preserved.

## Test plan

- Device / OS: iPad Mini gen 7, iPadOS 26.5.2
- Source media (container / video codec + profile / audio / HDR-DV / Subtitles): MKV, AVC, High 10, E-AC3, SDR, ASS (heavily stylized)
- Result: Playback successful, styles applied properly

*** 

- Device / OS: iPhone 17, iOS 26.5.2
- Source media (container / video codec + profile / audio / HDR-DV / Subtitles): MKV, AVC, High 10, E-AC3, SDR, ASS (heavily stylized)
- Result: Playback successful, styles applied properly

## Checklist

- [x] `CHANGELOG.md` updated


superuser404notfound:
Merged, thank you. The diagnosis is the valuable half here: getting from "some events are missing" to `appendLocked` treating the PTS as a unique key, by counting processed events against the source file, is what made this cheap to fix. The 245-against-2054 number tracking the 268 distinct timestamp pairs is about as clean a fingerprint as a report gets.

I pushed three follow-ups onto the branch before merging, so the diff that landed is not quite the one reviewed:

- **Order inside a shared PTS.** The insert used the lower bound of the run, so every new packet went in *ahead* of the ones harvested before it and a burst was stored in reverse. The store's array order carries through the drainer into `subtitleCues` (`insertCueSorted` keeps same-start text cues in arrival order), so a host renderer received overlapping lines in reverse authoring order. A new entry now joins the run at its end.
- **Two test fixtures.** `SubtitlePacketStoreTests.dedupOnRestartOverlap` and `Issue112PGSDisplaySetAssemblyTests.unarmedKeepsLegacyBehavior` both asserted the collapse using two *different* payloads on one PTS, which is exactly the case that now has to be retained, so both went red. A producer restart re-reads the same source bytes and its overlap is byte-identical, so the fixtures now deliver one packet twice, which is what they were describing all along.
- **The insert position.** It was scanned from the front of the array, making one append linear in retained packets and a session's harvest quadratic. That cost little while a burst collapsed to a single entry, and it is load-bearing now that a dense track keeps all of them: binary search instead.

`Issue235SamePTSRetentionTests` covers retention of a burst, collapse of a re-harvest inside a run, byte accounting on both paths, the ordering, backfill far behind the frontier, and the search boundaries. 1178 tests green, strict concurrency clean, tvOS and iOS Simulator builds green.

Cutting a release with this shortly and will follow up on #235 with the tag.

Issue #255, opened 2026-07-30
ChunkFetchDelegate reserves the whole file's Content-Length on a bodyless HEAD probe: malloc returns NULL and the force-unwrap kills the app on large sources


dlev02:
### What happened?

Opening a large progressive HTTP source crashes the app outright — `EXC_BREAKPOINT` / "Unexpectedly found nil while unwrapping an Optional value" — before a single frame plays. It reproduces on a 12.4 GB MKV; small files are fine.

The crashing thread is a URLSession delegate queue:

```
0  Foundation  Swift runtime failure: Unexpectedly found nil while unwrapping an Optional value
1  Foundation  __DataStorage.init(capacity:) + 284
2  Foundation  Data._Representation.reserveCapacity(_:) + 644
3  AetherEngine  specialized ChunkFetchDelegate.urlSession(_:dataTask:didReceive:completionHandler:)  (AVIOReader.swift:2215)
4  CFNetwork   __81-[__NSCFURLSessionDelegateWrapper dataTask:didReceiveResponse:completionHandler:]_block_invoke
```

`x0` is `0x0` in the crashed thread's register state — that is `malloc` returning NULL, and it's the value being force-unwrapped. `x19`/`x2` hold `0x318ffc000` (13,304,315,904 — 12.39 GiB), consistent with the requested capacity being the source's full `Content-Length`.

The cause is `ChunkFetchDelegate.urlSession(_:dataTask:didReceive:completionHandler:)`:

```swift
self.response = response
if let http = response as? HTTPURLResponse {
    let len = Int(http.expectedContentLength)
    if len > 0 { body.reserveCapacity(len) }
    ...
}
```

The reservation is whatever the origin declares, with no ceiling. `ChunkFetchDelegate` is reached through `syncRequest(_:budget:)`, and one of `syncRequest`'s callers is `headProbeFileSize()` — a `HEAD` request. A `HEAD` response carries `Content-Length: <entire file>` and delivers **zero body bytes**, so the delegate asks the allocator for the whole movie in order to buffer nothing. On a 6 GB device a 12.4 GiB `malloc` fails, and `__DataStorage.init(capacity:)` force-unwraps the NULL.

Two other threads in the same report were parked in `headProbeFileSize()` → `syncRequest` → `awaitSignal` waiting on exactly that response, which lines up: the source needed the HEAD fallback (the range probes hadn't yielded a length), and the response handler blew up on arrival.

[...]

PR #269, opened 2026-07-31
fix(hls): remux finite HEVC MPEG-TS VOD

### LINKED ISSUE #268: HEVC-in-MPEG-TS HLS VOD direct media playlist routes to AVPlayer with 0x0 video
### What happened?

A seekable remote HLS VOD whose **direct media playlist** contains HEVC video in MPEG-TS segments is routed to the native AVPlayer backend. AVPlayer accepts the item, exposes an AAC audio path and advances its clock, but AetherEngine reports no video dimensions and no video is presented.

This is the VOD/direct-media counterpart of #168, not a live-stream case:

- the playlist is a media playlist, not a master playlist;
- it contains `#EXT-X-ENDLIST`;
- its segments are MPEG-TS (`.ts`, no `#EXT-X-MAP`);
- the first actual segment probes as HEVC Main, 3840x2160, 25 fps, plus AAC;
- the source has a finite duration and must retain ordinary VOD seek semantics.

With `nativeRemoteHLS = false`, AetherEngine's normal source decision still ends at the native backend:

```text
aether_backend=native
aether_current_avplayer=present
duration_seconds=2698.760
video_width=0
video_height=0
phase=paused
error_code=aether_video_required
message="AetherEngine baseline requires video, but the source dimensions are 0x0"
```

The same source through the Hybrid AVKit surface confirms the misleading healthy-looking state:

```text
route=nativeAVPlayer
startup: currentTime 0 -> 2 s
seek: 2 -> 90 s, landing error 0
post-seek: currentTime 90 -> 92 s
```

The clock and seek advance, but the presented video remains black. This matches the carriage failure described in #168: AVPlayer does not build the HEVC video track from MPEG-TS HLS carriage.

### Why the existing fixes do not cover it

#168 shipped a TS -> fMP4 ingest reroute, but the current gate is live-only and depends on master-playlist variant evidence:

```swift
static func shouldArm(isLive: Bool, fallbackEnabled: Bool) -> Bool {
    isLive && fallbackEnabled
}
```

A direct media playlist has no `AVAsset.variants` advertisement, and this source has `isLive == false`, so that watchdog cannot arm.

#246 / 5.29.0 fixes second-open typed HLS detection and preserves the VOD resume anchor, but its destination remains the native remote-HLS bypass. That makes HLS admission reliable; it does not detect unsupported HEVC-in-MPEG-TS carriage before selecting AVPlayer.

The same live-only gate and VOD-native routing are still present in 6.3.0.

### Steps to reproduce

1. Serve a finite HLS media playlist with `#EXT-X-ENDLIST`.
2. Use MPEG-TS segments containing HEVC video and AAC audio.
3. Call:

```swift
let options = LoadOptions(
    httpHeaders: requiredHeaders,
    isLive: false,
    nativeRemoteHLS: false,
    autoplay: false
)

try await engine.load(
    url: mediaPlaylistURL,
    startPosition: nil,
    options: options
)
```

4. Observe that AetherEngine returns the native backend with `currentAVPlayer != nil`, duration is populated, but video dimensions remain `0x0`.
5. Starting playback advances media time and audio; seeking lands, but video remains black.

The private signed URL and credential-bearing headers are intentionally omitted. The playlist and first segment were fetched successfully immediately before the device run.

### Expected behavior

Once AetherEngine has content-backed evidence that a seekable HLS VOD carries HEVC in MPEG-TS, it should not hand that source directly to AVPlayer.

It should route the same URL and headers through a **seekable VOD remux path** that converts TS carriage to AVPlayer-compatible fMP4 while preserving:

- finite duration;
- `startPosition`;
- random forward/backward seek;
- source URL and HTTP headers;
- a single bounded load decision.

This must not use live semantics or `HLSLiveIngestReader`: treating the source as live would break the VOD seek contract.

### Suggested acceptance criteria

- Direct media playlists are supported; the decision does not require `AVAsset.variants`.
- HLS evidence comes from playlist/segment content, not the `.m3u8` suffix alone.
- HEVC-in-MPEG-TS VOD selects a seekable TS -> fMP4 remux route.
- URL, headers and `startPosition` are preserved.
- Forward and backward seeks land and continue advancing.
- H.264/AAC HLS VOD that AVPlayer supports can remain on the native path.
- Audio-only playlists are not mistaken for failed video.
- No retry through another source, provider or player.
- Tests cover both a direct media playlist and a master playlist carrying the same HEVC-in-TS VOD.

5.26.0 / `051fff80b5f4bd26981207b7a77426d4010792cb`

Also inspected through the 6.3.0 release source; the relevant ingest fallback remains live-only.

### Host app

SyncnextHybrid diagnostic smoke player

### Platform

tvOS physical Apple TV

### Playback path

Native AVPlayer selected by AetherEngine's normal routing decision

### Related issues

- #168 — same HEVC-in-MPEG-TS no-video-track signature, but fixed only for live master playlists
- #246 — reliable VOD HLS detection/reroute to the native bypass; does not address unsupported carriage

qoli:
Fixes #268.

## Problem

AVPlayer can mount finite HEVC-in-MPEG-TS HLS as audio-only / black video. This is a long-standing HEVC HLS playback-path problem, not a device-generation-specific issue.

The existing HLS reroute sends VOD back to native remote HLS. That is correct for ordinary HLS, but preserves the black-screen path for this carriage. Reusing the live ingest is also not suitable because its rolling-window contract breaks VOD seek semantics.

## Change

- classify the source from playlist and segment content before choosing the native route
- require `#EXT-X-ENDLIST`, MPEG-TS carriage, and PMT stream type `0x24` (HEVC)
- accept direct media playlists and the highest-bandwidth muxed master variant
- feed confirmed inputs through a bounded, ordered TS ingest
- expose duration and restart the finite playlist at the segment preceding a time seek
- preserve HTTP headers, resume position, forward/backward seek, and the existing demux/decoder/remux path
- leave H.264, fMP4, live, audio-only, encrypted/inconclusive, and unsupported playlist shapes on their existing route

This is content-gated; it does not infer codecs from the `.m3u8` suffix or retry from an AVPlayer error string.

## Validation

- `swift test --filter Issue268HLSVODIngestTests` — 4 passed
- `swift build -Xswiftc -strict-concurrency=complete` — passed
- `xcodebuild -scheme AetherEngine -destination 'generic/platform=tvOS' build CODE_SIGNING_ALLOWED=NO` — passed
- `swift test --skip-build --skip Issue220SoftwareDecoderDrainTests` — 1361 tests in 206 suites passed

A full unfiltered run has one failure in `Issue220SoftwareDecoderDrainTests` (expected drain counter >= 24, observed 17). The same failure reproduces in a clean worktree at upstream `83dca0b`, so it is not introduced by this branch.

The same reader/demux seam was physically validated in a downstream integration with finite duration, forward seek, post-seek progress, and 2x playback. This PR reconstructs that workaround against current upstream `main`; direct physical-device acceptance of this exact branch remains pending, hence the draft status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,15 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Fixed
+
+- **Finite HEVC-in-MPEG-TS HLS VOD no longer reaches AVPlayer's audio-only,
+  black native path.** A bounded content probe now confirms the playlist is
+  finite, its segments are MPEG-TS, and its PMT declares HEVC before selecting
+  a seekable TS-to-fMP4 ingest. Direct media playlists and master playlists are
+  both covered; URL, HTTP headers, duration, resume position, and forward or
+  backward seek semantics are preserved. H.264, fMP4, live, audio-only, and
+  inconclusive HLS inputs keep their existing route. (#268)
 
 ## [6.3.0] - 2026-07-31
 
--- a/Sources/AetherEngine/AetherEngine.swift
+++ b/Sources/AetherEngine/AetherEngine.swift
@@ -2397,6 +2397,26 @@ public final class AetherEngine: ObservableObject {
         // (audio-tap reader selection, seek paths) sees a genuine remote-HLS session.
         if RemoteHLSMediaSelection.shouldReroute(failure: probeFailure, isCustomSource: isCustomSource),
            case .url(let hlsURL) = source {
+            if let reader = try await HLSVODIngestReader.makeIfHEVCMPEGTS(
+                playlistURL: hlsURL,
+                httpHeaders: loadedOptions.httpHeaders
+            ) {
+                try checkLoadCurrent(gen)
+                EngineLog.emit(
+                    "[AetherEngine] AE#268: finite HEVC-in-MPEG-TS HLS; "
+                        + "routing through seekable TS -> fMP4 ingest",
+                    category: .engine
+                )
+                var remuxOptions = loadedOptions
+                remuxOptions.nativeRemoteHLS = false
+                return try await load(
+                    source: .custom(reader, formatHint: "mpegts"),
+                    startPosition: startPosition,
+                    options: remuxOptions,
+                    audioSourceStreamIndex: audioSourceStreamIndex,
+                    discTitleID: discTitleID
+                )
+            }
             EngineLog.emit("[AetherEngine] AE#154: HLS playlist on the VOD loopback path; rerouting to the native remote-HLS bypass", category: .engine)
             loadedOptions.nativeRemoteHLS = true
             do {
@@ -2924,6 +2944,26 @@ public final class AetherEngine: ObservableObject {
             if RemoteHLSMediaSelection.shouldReroute(failure: error, isCustomSource: isCustomSource),
                case .url(let hlsURL) = source,
                loadGeneration == gen {
+                if let reader = try await HLSVODIngestReader.makeIfHEVCMPEGTS(
+                    playlistURL: hlsURL,
+                    httpHeaders: loadedOptions.httpHeaders
+                ) {
+                    try checkLoadCurrent(gen)
+                    EngineLog.emit(
+                        "[AetherEngine] AE#268: second open confirmed finite HEVC-in-MPEG-TS HLS; "
+                            + "restarting through seekable TS -> fMP4 ingest",
+                        category: .engine
+                    )
+                    var remuxOptions = loadedOptions
+                    remuxOptions.nativeRemoteHLS = false
+                    return try await load(
+                        source: .custom(reader, formatHint: "mpegts"),
+                        startPosition: startPosition,
+                        options: remuxOptions,
+                        audioSourceStreamIndex: audioSourceStreamIndex,
+                        discTitleID: discTitleID
+                    )
+                }
                 EngineLog.emit(
                     "[AetherEngine] AE#246: the loopback session's own open classified the source as HLS; "
                     + "taking the AE#154 reroute onto the native remote-HLS bypass",
--- a/Sources/AetherEngine/Demuxer/CustomIOReaderBridge.swift
+++ b/Sources/AetherEngine/Demuxer/CustomIOReaderBridge.swift
@@ -15,6 +15,10 @@ final class CustomIOReaderBridge: AVIOProvider, @unchecked Sendable {
     private var isFullyClosed = false
     private(set) var isSeekable: Bool = true
 
+    var timeSeekableReader: TimeSeekableIOReader? {
+        reader as? TimeSeekableIOReader
+    }
+
     var cumulativeBytesFetched: Int64 { 0 }  // custom readers don't track network bytes
 
     /// #112 round 9: same demux-thread-only contract as AVIOReader's deadline. Armed by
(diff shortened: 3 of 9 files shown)


superuser404notfound:
Merged, thank you. The shape is right: the decision comes from the playlist and the PMT, the ingest is seekable, and every carriage AVPlayer handles itself stays on the native route. Follow-up hardening is on main in d8cc2734.

**One real defect, found before the merge.** `Demuxer.seek` hands the reader an ABSOLUTE source PTS, while the reader addresses elapsed playlist time. MPEG-TS carries an arbitrary PTS origin: ffmpeg writes 1.4 s by default, which is invisible because landing early is covered by the lower-bound contract, but broadcast-derived VOD routinely starts hours in. On a fixture muxed with `-output_ts_offset 1000`:

```
[HLSVODIngest]   seek target=1006.74s segment=98 origin=588.00s
[HLSVideoEngine] producer restarted at idx=1 (seek=1006.74s [absolute source-PTS])
[AetherEngine]   VOD shift published: 1586.701s (prev 1002.741s, delta 583.960s)
  t=30 state=playing cur=1592.60 src=1008.64 dur=600.0
```

A seek to item second 5 positioned the ingest at the last segment of a 600 s playlist, and the session clock left the item's own duration. `Demuxer` now strips the origin (the anchoring stream's `start_time`, else the container's) before positioning a `TimeSeekableIOReader`, and the protocol documents which axis it speaks. Same fixture afterwards: `seek elapsed=0.00s segment=0`.

Rest of the follow-up:

- `seekBounded` takes the same seam. It was still reaching `avformat_seek_file` (VOD cue prewarm, side reader positioning), which is exactly the byte binary search this source cannot serve. The prewarm now skips a segmented source outright: there is no index to load, and each reposition would refetch a segment for the uniform-stride plan it ends up building anyway.
- The carriage probe reads packet-aligned chunks and stops at its first verdict instead of accumulating up to 512 KiB byte by byte. That path runs on every HLS VOD reroute, including the H.264 sources that then stay native, where the per-byte loop measured ~75 ms of CPU per load against a loopback origin versus ~2 ms for a bulk read. The ceiling is 128 KiB now.
- `seek(offset: 0, whence: SEEK_SET)` no longer reports a rewind that never happened; only a reposition that does not move succeeds. As a side effect, the `isSeekable` line in `CustomIOReaderBridge` was a no-op before, because the old handshake already answered 0.
- The prefetch window narrows with the observed segment size, so four 4K segments in flight cannot peak near 100 MB (#255), and the FIFO matches the live ingest at 16 MB.
- `makeIndependentReader` returns nil rather than opening a second full HTTP ingest of the same VOD for side-demuxed subtitles or scrub previews.
- Tests for the restart mapping, the byte-seek refusal, the prefetch bound and the probe verdicts, and the two reroute call sites collapsed into one helper.

Verification here: full suite green (405 XCTest, 1365 Swift Testing in 207 suites), strict-concurrency build clean, tvOS BUILD SUCCEEDED, plus an end to end run through `aetherctl play` against a local range-capable origin. HEVC-in-TS VOD plays with duration 600.04 s, six forward and backward seeks land exactly in 52 to 100 ms, each restarting at the segment before its target, and the H.264 control keeps the AE#154 native route. `Scripts/fetch-fixtures.sh` now carries the fixture recipes (default origin, 1000 s origin, H.264 control) so the run is reproducible without a private source.

`Issue220SoftwareDecoderDrainTests` passed here on this branch, so that failure looks machine dependent rather than something the branch introduced.

The open piece is a confirmation on real hardware against the original source, so I reopened #268 for it.

Issue #98, opened 2026-07-03
Unify handling of dynamic range checks


DrHurt:
### What should the engine do?

All macOS 26+ devices and all iOS 26+ devices run in HDR mode by default on built in displays. Native apps & OS UI are all mastered in HDR. 

The OS transparently tone maps SDR-only apps and if needed, system can also physically switch displays to SDR which is instant and seamless as there is no HDMI handshake like TVs.

Intel Macs are SDR-only on internal displays and do the reverse with HDR apps.

Sodalite on iOS devices currently runs in SDR mode.

 The better alternative in my opinion would be to enable high EDR headroom in Sodalite (or the engine) via `.allowedDynamicRange(.high)` in the appropriate views before checking `currentEDRHeadroom` which should then return > 1 on HDR displays & == 1 on SDR displays.

This would create a simpler, unified dynamic-range detection algorithm across all supported devices and displays (including external ones)

Ideally this should be engine-side so it makes it to all host apps but will need testing.

Thanks again and good luck

### Motivating media or use case

Unified & simplified cross platform approach 

### Area

HDR / Dolby Vision

### Would you be willing to open a PR?

No

superuser404notfound:
Thanks for the detailed writeup, this prompted a proper look at all three detection paths.

Some context on where the engine already landed: since 4.10.0 the master-vs-media routing decision is a pure, tested function, and on iOS `AVPlayer.eligibleForHDRPlayback` counts as panel readiness (f9b6d55). HDR and DV sources on HDR-capable iPhones and iPads route through the master playlist since that release, so the SDR behavior you describe should be gone with a host build carrying 4.10.0 or newer. If you still see SDR playback on such a build, that is a bug and logs would be very welcome.

On the `currentEDRHeadroom` approach: we prototyped exactly that during the AE#88 retest before settling on eligibility, and reverted it for three reasons.

1. Ramp race. `currentEDRHeadroom` is a momentary value the system ramps while EDR content is on screen. Measured on an iPhone on iOS 26, it read 1.0 when launching playback from the SDR browsing UI. `.allowedDynamicRange(.high)` on views showing SDR artwork (which is all a Jellyfin browsing UI has) does not engage EDR by itself, so a load-time read races the ramp.
[...]

PR #280, opened 2026-08-02
fix(hls): break the live backpressure/blocking-reload deadlock

tschuegy:
## Summary

Live sessions freeze for 6-8 s at a time (picture + audio) on a repeating cadence. Root cause is a three-legged mutual wait between the segment producer's backpressure park and the LL-HLS blocking reload served by the loopback server:

1. `HLSSegmentProducer.advanceMuxer` parks once `head - bufferAheadSegments` exceeds the consumer's fetch high-water. The park releases **only** on a client segment GET (`declareTarget` from the serve paths).
2. Meanwhile AVPlayer's blocking reload asks for `_HLS_msn = head` — a segment only a producer cut can satisfy. `waitForLiveSegment` waits passively on `firstSegmentCondition`; there is no demand path back to the parked producer.
3. The held reload occupies the serialized HTTP/1.1 keep-alive connection (`handleConnection` loop), so AVPlayer's next segment GET — the one that would release the park — queues behind the hold.

The 18 s hold expires into `503 blocking reload msn=N unsatisfiable` long after AVPlayer's ~4 s forward buffer (`preferredForwardBufferDuration` default on the loopback path) has drained into `playbackStalled`. Recovery is then the generic stall watchdog (nudge at +6 s, item reload at +12 s) — the visible multi-second freeze. The live pump watchdogs cannot fire during the park because the pump is blocked inside `advanceMuxer`, below the top of `readLoop`, and the VOD wedge breaker is `!isLive`-gated.

Log signature (Apple TV 4K, tvOS 26, Xtream MPEG-TS 1080i/p25 H.264):

```
[LagDiag] ... fwd=5.2 ... fwd=1.2 (5 s without a single segment fetch)
[NativeAVPlayerHost] #1 playbackStalled
[HLSLocalServer] -> 503 /media.m3u8 reason=blocking reload msn=24 unsatisfiable
[HLSSegmentProducer] #65 backpressure released (advance) head=24 target=14 after=18s
[AetherEngine] #65 re-engaging stalled AVPlayer (stall + 6s without fetches): nudge seek ...
```

## Changes

- **Live is source-paced: the advance and versioned-init parks are now VOD-only.** Overproduction on live is bounded by the origin's real-time delivery; the only unbounded case is a consumer that stopped polling entirely, which a new logged runaway guard (`awaitLiveWindowHeadroom`, 180 resident segments) catches. Unlike the park it releases on window eviction, not consumer fetches, so it cannot deadlock against a held reload. It also logs from the first cycle — the old live park was silent below 12 s, which is why the consumer-facing freezes never showed a producer-side line.
- **`LiveWindowSizing.windowSegmentCount` is sized by the observed segment cadence** (rolling mean EXTINF of the last 20 finalized segments), not the cut target alone. Under `.fastZap` the 0.5 s cut target vs ~2 s GOPs inflated the "60 s" window to 120 segments (~240 s): `EXT-X-MEDIA-SEQUENCE` stayed pinned at 0 and `evictBelow` never fired in short sessions. The divisor is `max(cutTarget, observed)`, so sub-target keyframe bursts cannot widen the window; `minSafeSegments` still floors it.
- **`reloadStalledConsumerItem` honors `LiveReloadPolicy`** — live rejoin with `startPosition: nil` + `skipInitialSeek`, same contract as the #98 media fallback and the reload paths in `AetherEngine+Loading`. Previously it issued the zero-tolerance seek into a possibly-slid window that `LiveReloadPolicy`'s doc comment forbids for live rejoins.
- **The blocking-reload hold is bounded by `3 x` the sealed TARGETDURATION** (= the advertised HOLD-BACK) via a new `liveBlockingReloadHoldSeconds` protocol member (default extension keeps 18.0 for other conformers). `.standard` stays exactly 18 s; `.fastZap` (TD=2) drops to 6 s — a hold that outlives the consumer's forward buffer guarantees the stall it exists to prevent.

VOD paths are byte-identical: both parks, the wedge breaker, and the VOD window math are untouched.

## Test plan

- **Device:** Apple TV 4K (3rd gen, A15), tvOS 26.6, live MPEG-TS over Xtream (H.264 1080p25/1080i25, AAC + MP2 audio, ~2 s GOPs, `.fastZap` join profile, `max_connections=1` origin). Pre-fix the freeze reproduced roughly every 2 minutes; post-fix, initial sessions play clean — no `503 ... unsatisfiable`, no `backpressure released (advance)`, no `playbackStalled` in the logs, and `EXT-X-MEDIA-SEQUENCE` starts sliding after ~60 s. A longer soak is still running; I will report back on this PR if anything surfaces.
- **Unit:** full suite green on this branch rebased onto current `main` (1411 tests), including 6 new tests (`LiveWindowObservedCadenceTests`) pinning the observed-cadence window math, the slide behavior, and the hold bound (unsealed fallback 18 s, sealed TD=2 -> 6 s, seal stability).
- `CHANGELOG.md` updated under `[Unreleased]`.

## Notes

- The new runaway-guard cap (180 resident segments) is intentionally far above the steady-state window; if it ever logs `live headroom PARK`, the consumer has stopped polling and the engine's existing 12 s watchdogs are already the real recovery. Happy to rename/tune per your conventions.
- Host apps that want the old behavior back can keep forcing `LoadOptions.liveBlockingReload = false`; with this fix the blocking-reload contract is actually satisfiable again, so we left the engine default untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,24 @@ the public-API contract.
   to an unobservable DV panel: it reads the same attribution the settle branch
   got in #274, so an engine rate-only write that never reports an end says so
   rather than claiming DV. Measured from the logs on Sodalite#49.
+- **Live sessions no longer freeze 6-8 s at a time when the producer's
+  backpressure park meets an LL-HLS blocking reload.** The advance park
+  released only on a client segment GET, while the client's held
+  `?_HLS_msn=` reload was only satisfiable by a producer cut — and the held
+  reload occupies the serialized keep-alive connection, starving the very
+  segment GET that would release the park. The 18 s hold then expired into
+  `503 unsatisfiable` long after AVPlayer's ~4 s forward buffer had drained
+  into `playbackStalled`. Live production is source-paced now: the advance
+  and versioned-init parks are VOD-only, replaced on live by a logged
+  resident-segment runaway guard that releases on window eviction instead of
+  consumer fetches. Three aggravators fixed alongside: the sliding window is
+  sized by the observed segment cadence instead of the cut target (fastZap's
+  0.5 s target vs ~2 s GOPs inflated the window 4x, pinning MEDIA-SEQUENCE
+  at 0 and deferring `evictBelow` for minutes), the stall-recovery item
+  reload now honors `LiveReloadPolicy` (live rejoin: no stale-clock resume,
+  no zero-tolerance initial seek), and the blocking-reload hold is bounded
+  by `3 x` the sealed TARGETDURATION (= the advertised HOLD-BACK) instead of
+  a hardcoded 18 s. VOD paths are byte-identical.
 
 ## [6.4.4] - 2026-08-02
 
--- a/Sources/AetherEngine/AetherEngine.swift
+++ b/Sources/AetherEngine/AetherEngine.swift
@@ -1679,14 +1679,20 @@ public final class AetherEngine: ObservableObject {
         stallRecoveryWindowUntil = Date().addingTimeInterval(Self.stallRecoveryWindowSeconds)
         EngineLog.emit(
             "[AetherEngine] #65 nudge did not revive the consumer; reloading item at "
-            + "\(String(format: "%.2f", anchor))s"
+            + (isLive ? "the live edge (rejoin)" : "\(String(format: "%.2f", anchor))s")
             + Self.recoveryAnchorLogSuffix(
                 anchor: anchor, position: position,
                 pendingSeekTarget: pendingRecoverySeekClockTarget)
             + " (same URL, same host)",
             category: .engine
         )
-        host.load(url: url, startPosition: anchor, inPlaceSwap: true)
+        // Live reload = live REJOIN: no stale-clock resume, no explicit start seek — the
+        // zero-tolerance seek into a possibly-slid window wedges the fresh item in waitingToPlay.
+        // Same contract as the #98 media fallback above; see LiveReloadPolicy.
+        host.load(url: url,
+                  startPosition: isLive ? nil : anchor,
+                  skipInitialSeek: LiveReloadPolicy.skipInitialSeek(isLive: isLive, isRejoin: true),
+                  inPlaceSwap: true)
         host.play()
         if let ordinal = nativeSubtitleReapplyOrdinal {
             EngineLog.emit(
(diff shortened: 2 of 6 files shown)


superuser404notfound:
Merged as b9b803dc. Thanks for this one. The report carried its own proof: the log signature, the three-legged wait, and the observation that the live park is silent below 12 s all hold up against the tree, and that last detail is why this never surfaced in an earlier live report.

Verified before merging:

- `awaitBackpressureRelease` releases only through `cache.awaitFetchHighWater`, and `currentTargetIndex` advances only in `declareTarget` (`SegmentCache.swift:221`), so a client segment GET really is the sole release. Same half as AE#65 Piece A.
- Every breaker on that park is `!isLive`-gated (`HLSSegmentProducer.swift:1059` and `:1072`), so live had neither a breaker nor a log line.
- Dropping `awaitPrefetchDiskBudgetRelease` from the live branch is an exact no-op: the retention budget is hard 0 for live sessions (`HLSVideoEngine.swift:1041`), and the park returns immediately at 0.

Local gates: 1424 tests green (the branch was one release behind main, hence 1411 there), `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` clean, tvOS Simulator build succeeded.

One correction pushed on top in 4ae50e2b. The doc comment on `awaitLiveWindowHeadroom` says the guard releases on eviction (window slide) rather than on a consumer fetch, and therefore cannot deadlock against a held blocking reload. The first half does not hold while the park is active: a parked pump finalizes no segment, so `segments.count` stops growing, `newFirst` stops advancing, and `notePlaylistBuild -> evictBelow` evicts nothing. While the park holds, the only thing that lowers `cache.count` is `pruneOutsideWindow` off a `declareTarget`, which is exactly the consumer segment GET the old park waited on.

That does not make the fix wrong, it makes the reason it works a different one, and the difference is load-bearing for whoever touches the constant next: what keeps the deadlock out of reach is the height of the cap, not the release path. 180 resident segments is only reachable by a consumer that has already stopped, and the 12 s stall watchdog reloads the item (issuing a fresh GET) well before that. Lowering the cap toward the steady-state window would put the deadlock back in reach, so the comment says so now.

The same commit adds a latched diagnostic for the one case the removed advance park used to make impossible. That park capped the producer 10 segments ahead of the consumer's fetch point, so the sliding window could never overtake it. Source-paced live cannot get there, as the PR argues, but an origin that hands over more than one window of backlog faster than the consumer drains it can, and the consumer then asks for a segment `evictBelow` has already deleted. Downstream that reads as a cache miss or a live-edge jump with nothing naming the cause, so `notePlaylistBuild` now logs `live window slid past the consumer` once per excursion. No behavior change, just something to grep for.

Interested in the soak result whenever it lands, in particular whether `EXT-X-MEDIA-SEQUENCE` keeps sliding past the first few minutes, and whether `live headroom PARK` ever appears at all.

Issue #281, opened 2026-08-02
VOD cold-start pays a mandatory sequential head→tail→re-anchor round trip on non-fast-start MP4 sources — request for parallel fetch or moov-offset hint


digilearn-dev:
### What happened?

On every fresh VOD load against a non-fast-start MP4 source (moov atom at file end), the demuxer open sequence costs 1.2-2.8s of pure network round-trip time before frame rate / stream metadata is available, via three sequential connections:

gen=1: offset=0, len=32MB           → real data connection (~1-2s)
gen=2: offset=<near file size>, len=0MB  → moov atom / tail probe (~130-750ms)
gen=3: offset=<near media start>, len=32MB → re-anchor (~45-100ms)

This is confirmed consistent across multiple sessions and titles on our end. It's occurring before any codec/DisplayCriteria logic engages, so it's purely the cost of locating and reading the trailing moov atom.

I checked #69 and see this matches the documented non-fast-start MP4 behavior — the detour-block cache fix there collapsed what could otherwise be a reconnect storm down to this legitimate minimum sequence, but the sequential round-trip cost itself remains by design. I'm not reporting a regression from that fix; I'm asking whether there's room to reduce this legitimate minimum further, since for slower/higher-latency sources (debrid-style proxies in our case) this is now the single largest contributor to VOD time-to-first-frame — bigger than anything in the DisplayCriteria path (#274/#277).

### Steps to reproduce

1. Load a VOD title from a non-fast-start remote MP4 source (moov atom at end of file) with meaningful network latency to the host (not a fast local/CDN path).
2. Observe AVIOReader logs: pump conn start gen=1 offset=0, followed by pump conn start gen=2 offset=<near file size> len=0MB, followed by pump conn start gen=3 offset=<near 0>.
3. Measure elapsed time from open_started to the point frame rate/stream info becomes available — the three round trips dominate this window on higher-latency sources.

### AetherEngine version or commit SHA

6.4.3

### Host app

Custom / my own integration

### Platform

tvOS

### OS version

tvOS 26.6 (beta)

### Device / chip

Apple TV 4K (A15)

### Playback path

Native AVPlayer

### Source media (for playback bugs)

Various VOD titles served via a remote HTTP source (debrid-style proxy) producing non-fast-start MP4 (moov atom at file end). Not codec/resolution-specific.

### Error codes / log lines

```text
[AVIOReader] pump conn start gen=1 offset=0 len=32MB
[...]

PR #285, opened 2026-08-02
fix(io): stop re-fetching bytes the open already has

superuser404notfound:
Follow-up to #281, from measuring rather than reasoning. aetherctl against a Range-logging origin, four MP4 layouts, all sized above the 32 MB range so their tails lie outside the first window.

## Three defects, all the same shape

**A range that finished delivering was fetched again.** A completed range clears `activeTask` exactly as a dropped one does, and the no-connection branch reconnected at the READ position regardless, which resets `winStart` and drops everything still resident. The parse pass is precisely the consumer this bites: it reads one 256 KB AVIO buffer at a time while the transfer runs at line rate. A 764450 B trailing `moov` cost three connections and 1506918 delivered bytes, 1.97x its own size, each reconnect landing exactly one AVIO buffer further on. The branch already excluded the hard-cap end for this reason; the reasoning was never extended to a range that simply finished. Serving what is in hand first also revives the #220 frontier refill, which could not run while this branch preempted it on every completed range.

**A read at EOF opened a connection for it.** The EOF decision sat below that reconnect, so a position at exactly `fileSize` issued `bytes=<fileSize>-` and took an empty 206, whose reconnect reset `winStart` past the last byte and dropped the window the parse was still reading.

**The parked window was cut from the wrong place.** #281 parks at seek time from `winStart`, on the reasoning that the demuxer returns to the window's start. It returns to the FILE's start: landings of 48, 1161, 5752 here and 265159 in the field trace on #281. Those coincide only when the parse seeks away before reading anything. A fragmented MP4 reads 33 MB first, so the parked copy begins megabytes past everything the return asks for and covered nothing. The head is now collected as the data connection delivers it, which is the only point at which it can be: `trimWindowLocked` drops it as the parse moves forward.

## Measured

Request counts, loopback:

| layout | before | after |
|---|---|---|
| moov at end, 764450 B | 5 | 3 |
| fragmented, mfra at end | 5 | 3 |
| moov at end, 23337 B | 2 | 2 |
| faststart | 2 | 2 |

Wall clock, same origin with 300 ms of latency per request (loopback cannot show an overlap):

| layout | before | after |
|---|---|---|
| fragmented, mfra at end | 1732 ms / 4 req | **1219 ms / 3 req** |
| moov at end, 764450 B | 678 ms / 3 req | 708 ms / 3 req |
| moov at end, 23337 B | 390 ms / 2 req | 385 ms / 2 req |
| faststart | 420 ms / 2 req | 420 ms / 2 req |

## Tests

Three new, each verified red against the old behaviour: a completed range is read out of the window, a read at EOF opens nothing, and the return to the head survives a parse that read past the window's start. 1435 green.

## Kept on the argument rather than on evidence

The parked window stays. It covers a return to a mid-file offset, which the retained head cannot, and it did not fire on any layout measured here.

https://claude.ai/code/session_01HX5zV3Fcf7Nzq4DYGdXvQE

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,25 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Fixed
+
+- **A range that finished delivering is now read out of the window instead of
+  fetched again.** A completed range clears `activeTask` exactly as a dropped
+  one does, and the no-connection branch reconnected at the READ position
+  regardless, which resets `winStart` and drops everything still resident. A
+  consumer slower than the transfer, which is what the parse pass is (one 256 KB
+  AVIO buffer at a time), therefore re-fetched what it had just been handed.
+  Measured with aetherctl against a Range-logging origin: a 764450 B trailing
+  `moov` cost three connections and 1506918 delivered bytes, 1.97x its own size.
+  It now costs one. Serving what is in hand first also lets the #220 frontier
+  refill run, which it could not while this branch preempted it on every
+  completed range.
+
+- **A read at the end of the file no longer opens a connection for it.** The EOF
+  decision sat below the reconnect, so a position at exactly `fileSize` first
+  issued `bytes=<fileSize>-` and took an empty 206 whose reconnect reset
+  `winStart` past the last byte, dropping a window the parse was still reading.
+  On the same trailing-`moov` measurement that was one of the three connections.
 
 ## [6.5.1] - 2026-08-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,8 @@ the public-API contract.
   is the only point at which it can be, since `trimWindowLocked` drops it as the
   parse moves forward. Measured with aetherctl against an origin with 300 ms of
   latency per request: opening a fragmented fixture went from 1732 ms and four
-  requests to 1219 ms and three.
+  requests to 1219 ms and three. The parked window is gone: across six container
+  layouts (four MP4, two MKV) it served no read that the retained head does not.
 
 ## [6.5.1] - 2026-08-02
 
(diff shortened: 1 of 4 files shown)


superuser404notfound:
Both points I had left on an argument rather than on a measurement are now measured, and one of them did not survive.

## The parked window is gone

It served no read on any of six container layouts. The demuxer returns to the head of the FILE in every one of them, which the retained head covers, so the park was carrying a premise the measurement contradicts plus up to 4 MB of transient footprint next to the copy that actually answers. Request counts are unchanged by the removal on all six.

| layout | requests | what answered |
|---|---|---|
| faststart.mp4 | 2 | nothing |
| moovend.mp4 | 2 | tail prefetch (trailing `moov`, 23337 B) |
| bigmoov_end.mp4 | 3 | retained head (trailing `moov`, 764450 B, too big to fit the suffix) |
| frag.mp4 | 3 | tail prefetch (`mfra`) + retained head |
| cues_end.mkv | 2 | nothing |
| cues_front.mkv | 2 | nothing |

Worth recording separately: matroska does not read its Cues at open, wherever they sit. The speculative fetch is spent on both MKV layouts.

## The speculative suffix fetch stays, and now there is a number for it

It pays off on two of the six, so the question was whether the other four are paying for them. Five runs per point at 250 KB/s with 300 ms of latency, spread under 30 ms:

| layout | with | without | delta |
|---|---|---|---|
| faststart.mp4 | 1587 ms | 1335 ms | **+252 ms** |
| cues_end.mkv | 1590 ms | 1350 ms | **+240 ms** |
| moovend.mp4 | 1716 ms | 2076 ms | **-360 ms** |

The cost is exactly the link time of 64 KB (64 KB / 250 KB/s = 256 ms). The benefit is one round trip plus the sequential tail fetch it replaces. They are equal at about 2 Mbit/s, which cannot stream the material this engine opens; at 10 Mbit/s the cost is 51 ms against a round trip the #281 field trace measured at 438 ms. Keeping it.

1435 tests green.

PR #301, opened 2026-08-04
fix(io): let custom readers skip disc probing

murderer1234:
## Summary

- add `IOReader.discImageProbeEnabled`, defaulting to `true`
- skip `DiscReader.wrap` when a custom reader declares that it represents ordinary media
- expose the same opt-out on `SMBIOReader` and preserve it across independent-reader clones
- document the behavior and cover both the opt-out and compatibility default with regression tests

## Why

Every custom `IOReader` currently enters disc recognition before its supplied container hint is applied. For an ordinary remote media file, that performs sparse reads at the ISO9660 and UDF signature offsets even though the caller already knows the source is not a disc image. Those reads are network round trips, and the cost can recur when playback creates independent readers for subtitles or frame extraction.

URL sources already gate remote disc recognition by file extension. Custom readers have no equivalent way to communicate that source fact.

## Compatibility

The new property defaults to `true`, so existing readers and raw ISO/UDF sources keep automatic DVD/Blu-ray recognition. The public `MediaSource.custom` case remains unchanged, avoiding a source break for callers that pattern-match its associated values.

## Validation

- `swift test`
- 1493 tests passed, 0 failures
- regression coverage verifies that opting out avoids seeks to `0x8001` and `256 * 2048`
- SMB coverage verifies that the default remains enabled and independent readers preserve an explicit opt-out

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,15 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Fixed
+
+- **Ordinary remote custom readers can skip ISO/UDF disc-image recognition.**
+  `IOReader.discImageProbeEnabled` defaults to `true`, preserving automatic DVD
+  and Blu-ray image support. Readers that already know they expose a regular
+  media file can return `false`, avoiding the sparse remote seeks used only for
+  ISO9660 and UDF signature checks. The policy travels with independent readers,
+  so subtitle side demuxers, reloads, and frame extraction do not repeat those
+  unnecessary network reads.
 
 ## [6.6.3] - 2026-08-04
 
--- a/README.md
+++ b/README.md
@@ -242,9 +242,13 @@ final class MyArchiveReader: IOReader {
     func seek(offset: Int64, whence: Int32) -> Int64 { /* ... */ }  // AVSEEK_SIZE (65536) returns total size
     func close() { /* ... */ }
 
-    // Optional (both have defaults). Override to unlock extra features:
+    // Optional requirements have defaults. Override to unlock extra features:
     func cancel() { /* unblock a blocked read at teardown, do NOT invalidate the reader */ }
     func makeIndependentReader() -> IOReader? { /* a fresh cursor over the same source, or nil */ }
+
+    // Optional, defaults to true. Return false when this is known to be an
+    // ordinary media file rather than a raw ISO/UDF disc image.
+    var discImageProbeEnabled: Bool { false }
 }
 
 let probe = try await engine.load(source: .custom(MyArchiveReader(), formatHint: "mp4"))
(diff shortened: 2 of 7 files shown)


superuser404notfound:
Thanks for this, and for the thorough write-up. I reviewed and verified it locally:

- `swift build` clean in Swift 6 language mode, `swift test` green (1493 tests, 222 suites, plus the XCTest targets).
- I checked that the new regression test actually encodes the case: with the gate removed, `CustomDiscProbePolicyTests` records seek offsets `[32769, 524288, 0]` and fails on both assertions, so it is a real witness and not a green-by-construction test.
- The default keeps ISO/UDF recognition for every existing conformer, and the requirement is declared on the protocol rather than only in the extension, so a host override actually dispatches through the `any IOReader` existential. That detail bit us once before with `cancel()`, good that it is right here.

The reasoning holds for custom readers in particular: unlike a URL source there is no path extension to gate on (custom sources share a placeholder URL internally), so an explicit source fact from the caller is the only reliable signal.

One optional nit while it is still in draft: the README SMB snippet now shows `discImageProbeEnabled: false` in the canonical example. That is correct for the `.mkv` path in the sample, but people copy examples verbatim, and a copied `false` would silently disable disc support for someone pointing `SMBIOReader` at an ISO on the same share. A short comment on that line, or keeping the plain call and showing the opt-out separately, would avoid it.

Nothing blocking from my side. Let me know when you want it out of draft and I will merge it.

murderer1234:
Thanks for the review and for verifying the witness. I applied the README suggestion in ab84715: the canonical SMB example now keeps the safe default, and the ordinary-media opt-out is documented separately with an explicit note to retain the default for raw ISO/UDF images. The follow-up is docs-only and git diff --check passes. I have marked the PR ready for review.

superuser404notfound:
Merged as `72d68245`. Thanks for this, and for turning the README note around so quickly: the canonical SMB example keeping the safe default with the opt-out documented beside it is exactly right, and better than what I suggested.

One thing I resolved by hand on the way in, so it does not surprise you: your CHANGELOG entry sat under `[Unreleased]` at a point where that section still held what has since gone out as 6.6.4. It moved down beside the two entries that are now waiting there, unchanged in wording.

The next release carries it, and both consumer apps get it in the same pass.

There are three follow-ups I noted while reviewing, all engine-side and none of them yours to carry:

1. `DiscRecognitionCache` memoizes only successful recognitions, so a reader that does not opt out still repeats the two sniff reads on every open. The negative verdict is cacheable too, carefully: a transient read failure must not be frozen into "not a disc" for the session.
2. `Demuxer.open(reader:)` runs disc recognition even for `isLive: true`, while the URL path in `openHTTP` excludes live. `HLSLiveIngestReader` refuses every seek so it costs nothing there, but a host live reader that does seek would make two real reads at 0x8001 and 512K in the middle of a live join.
3. `FrameDecodeContext` and the probe open pass no `discCacheKey` at all, so neither can hit the positive cache in the first place.

Your opt-out is the right mechanism regardless of those: a custom source has no path extension to gate on, unlike a URL, so an explicit fact from the caller is the only reliable signal.

Issue #169, opened 2026-07-21
Loopback-HLS: the final segment of a long 4K DV title never completes at EOM — AVPlayer parks ~0.1s from the end (WaitingToMinimizeStalls) → -12889; resuming into the tail can't start a master


rrgomes:
On the loopback-HLS native path, a long, heavy **4K Dolby Vision P8.1** direct-play title fails to finish: the **final segment doesn't serve**, so AVPlayer parks a fraction of a second from the end and eventually errors instead of reaching `didPlayToEndTime`. A resume into the tail region then can't start a master at all.

**Asset:** 48-min episode, HEVC **Main 10 Dolby Vision Profile 8.1** (BL compat id 1), 24p, ~25.4 Mbit/s, EAC3 5.1 + Atmos, direct-play `.mkv` over the loopback path. `engine.duration` = 2881.3 s.

**Symptom 1 — final segment never completes (initial playthrough).** AVPlayer reaches the last segment and parks:

```
engineClock=2881.202  avpClock=2881.202  tcs=1 rate=1.0
waitReason=AVPlayerWaitingToMinimizeStallsReason
seekable=[0.0-2881.3]  loaded=[2844.8-2881.3]        <- loaded right to the end, but not playing out
```

It never fires `didPlayToEndTime`; after ~43 s it dies with **`CoreMediaErrorDomain -12889`** (loaded ranges collapse to `[]`).

**Symptom 2 — resume into the tail can't start a master.** Resuming at ~2878.5 s (the last ~3 s), the master never produces tracks and degrades all the way to the media playlist:

```
[HLSLocalServer] seg719: slow serve, sending early chunked header      (repeats)
#35 readiness gate: master did not start (timedOut) after a panel switch; reloading the master (attempt 2/2) at 2878.50s
#35 readiness gate: master never produced tracks; trying the HDR-preserving reduced master (DV dropped) at 2878.50s
#35 readiness gate: master never produced tracks after 2 attempts; falling back to the media playlist (HDR10 base, DV dropped) at 2878.50s
```

then AVPlayer stalls again (`WaitingToMinimizeStalls`, `loaded=[2862.7-2877.5]`).

[...]

Issue #71, opened 2026-06-26
AVIOReader reconnect loop under sustained 429: seek-driven reconnects reset the give-up streak, so a rate-limited origin never fails cleanly


ghost:
### What happened?

What happened?
When an origin rate-limits (returns 429 on essentially every request), AVIOReader keeps reconnecting instead of backing off and failing cleanly. Two interacting causes: (1) the 429 responses carry no Retry-After, so backoffBeforeReconnect gets retryAfter=0; (2) random-access parse reads call seekReconnect, which sets unproductiveReconnects = 0 — because the parser alternates seeks, the give-up streak is reset before it reaches the cap in recordReconnectAndShouldGiveUp, so the reader loops indefinitely (gen=N climbs) rather than surfacing a clean failure. Expected: on 429/503, honor backoff and give up cleanly after a bounded number of attempts.

### Steps to reproduce

Steps to reproduce
1. Stream a remote file from an origin that is actively rate-limiting (or generate the load by opening several large range-heavy files in quick succession against the same host).
2. engine.load(url:).
3. Watch the [AVIOReader] logs: status=429, reconnecting (streak=…) and chunk fetch got HTTP 429, with gen=N climbing and no clean termination.

### AetherEngine version or commit SHA

3.13.2

### Host app

Custom / my own integration

### Platform

macOS

### OS version

27 beta

### Device / chip

m2 / a9 pro

### Playback path

Native AVPlayer

### Source media (for playback bugs)

Any remote HTTP(S) source; IO-layer failure independent of container/codec. Most visible on a non-faststart / coarsely-interleaved MP4 (more parse seeks → more churn).

### Error codes / log lines

```text
[AVIOReader] chunk fetch got HTTP 429 at offset 0 (cached URL, will retry source)
[AVIOReader] chunk fetch got HTTP 429 at offset 0
[AVIOReader] Persistent conn ended at offset … status=429, reconnecting (streak=1 retryAfter=0.0s)
[AVIOReader] Persistent conn ended at offset … status=429, reconnecting (streak=2 retryAfter=0.0s)
[AVIOReader] Persistent conn start gen=15 offset=…      # gen keeps climbing, no give-up
```

### Anything else

[...]

Issue #12, opened 2026-05-23
Cache resolved redirect URL for subsequent HTTP range requests


Delarkz:
## Summary

It would be useful if AetherEngine remembered the final URL after an HTTP redirect during a media load and reused that resolved URL for later range requests, seeks, reconnects, or pipeline rebuilds.

For debrid/proxy playback URLs, the app often starts with a stable-looking playback endpoint that redirects to a CDN/object-storage URL. If every new HTTP request starts again from the original playback URL, playback can pay an extra proxy request and redirect hop repeatedly.

## Proposed Behavior

When opening an HTTP(S) media URL:

1. Start with the caller-provided source URL.
2. If the request redirects, keep the final resolved URL in memory for that media load.
3. Use the resolved URL for subsequent range requests, reconnects, and seeks.
4. If the resolved URL fails with an expiry-like status such as `401`, `403`, `404`, `410`, or a CDN-specific auth error, fall back to the original source URL and allow it to redirect again.

## Safety / Correctness Notes

- Keep the original URL as the canonical fallback because signed CDN URLs can expire.
- Preserve `Range` and `Accept-Encoding: identity` behavior across redirects.
- Consider exposing this as internal transport behavior rather than a public API requirement.

## Expected Benefit

This should reduce startup/seek/reconnect latency for redirected playback URLs and avoid repeated debrid/proxy hops while keeping the original source URL available when the resolved target expires.

superuser404notfound:
Implemented in `601205e`. After the first probe / chunk fetch completes with a 2xx/206, `AVIOReader` captures `dataTask.currentRequest?.url` (the post-redirect URL) and reuses it for every subsequent Range request. Cache is per playback session (one `AVIOReader` per `Demuxer`).

Safety rails:
- Cache only populates on success, so 4xx redirect targets can't poison it.
- 401 / 403 / 404 / 410 against the cached URL drop the cache and trigger a one-shot retry against the original source URL so the proxy can re-issue a fresh signed redirect. `fetchChunk` wraps the request in a two-pass helper that handles the fallback.
- `Range` and the caller-supplied extra headers are re-applied on cross-host redirects in `ChunkFetchDelegate` (the probe path already did this).

Scope: seekable mode only. Streaming mode is a single GET, so caching doesn't help there.

[...]

PR #308, opened 2026-08-05
fix(avio): bound reconnects for error-ended zero-byte connections

### LINKED ISSUE #307: AVIOReader: unbounded reconnect spiral (zero backoff) when the origin refuses a range refill; pinned redirect URL never invalidated on 5xx
**Media:** 3.7 GB H.264 (avc1.640033, 3840x1920) + AAC MKV, VOD, served by an Xtream-style aggregator with `max_connections=1`, `302→http` redirects, and no suffix-range support. Apple TV 4K, tvOS 26. AetherEngine 6.7.0, native dispatch path.

**Symptom:** Playback freezes ~1 min in, then dies. The log shows the reader wedged at a 32 MiB range frontier, reconnecting ~15×/s with zero backoff — 925 reconnects in 60 s — until the segment provider's forward-wait escalation tears the demuxer down (`demuxer.readPacket threw: read failed (-1)` → `pump finished: reason=readError(-1)`), after which the session is unrecoverable:

```
[AVIOReader] pump conn start gen=5 offset=131554695 len=32MB
[AVIOReader] pump conn gen=5 ended with error: cancelled
[AVIOReader] pump conn start gen=6 offset=131554695 len=32MB
[AVIOReader] pump conn gen=6 ended with error: cancelled
... (925 generations, same offset) ...
[AVIOReader] slow read: 60012ms at offset=131554695 detour=0(0ms,0fetch)
  stallWaits=925(59365ms,925signaled) reconnects=925 backoff=0ms gen=5->931
[HLSSegmentProducer] pump finished: reason=readError(-1)
```

**Root cause (three stacked defects):**

1. A connection that ends in **error with zero bytes delivered** for its generation lands in the no-connection reposition branch (`AVIOReader.swift` `readPersistent`, the `activeTask == nil, !connEndedByBackpressure, !windowCanServe` guard). That branch reconnects via `seekReconnect`, which **resets `unproductiveReconnects` and applies no backoff**, so the ended-connection ladder (status accounting, Retry-After, backoff, `.reconnecting`, bounded give-up) is unreachable. The spiral is unbounded by design.
2. The rejected response's status is **never logged** (only the 200-ignored-Range case is), so the storm is unexplained in the field log.
3. The **pinned post-redirect URL** is only invalidated on 401/403/404/410. An aggregator whose redirect targets expire per connection answers every later range with a hard 500 from the pinned URL, and the reader never falls back to the source URL for a fresh redirect.

Bonus: once the pump dies, `handleVODReadErrorExit`'s revive cap (#169) exhausts **silently** — `onVODSourceFailed` is only fired from the produced-nothing arm, so the host sees a forever-stalled AVPlayer with no error.

**Repro:** loopback origin that 302-redirects to a "CDN" which serves the first bounded range and answers 500 for every request at the next frontier (deterministic repro used for the tests in the linked PR; also reproduced with `aetherctl play` against such an origin).

**Expected:** an error-ended zero-byte connection takes the failure ladder (backoff, bounded give-up), a hard 5xx from a pinned URL drops the pin and re-resolves through the source URL, and revive-cap exhaustion surfaces `onVODSourceFailed`.

PR with fix + tests follows.

tschuegy:
Fixes #307.

A connection that ended in error without delivering a single byte of its generation was routed through the no-connection reposition branch, whose `seekReconnect` clears the unproductive streak and applies no backoff. An origin refusing every refill (connection-capped Xtream panel answering 500 at a 32 MiB range boundary) therefore produced an unbounded reconnect spiral — ~15 connections/s, 925 reconnects in 60 s observed in the field — until the segment provider's forward-wait escalation tore the demuxer down and killed the session.

**Changes**

- `AVIOReader.readPersistent`: exclude error-ended zero-byte connections (`connEnded && !connEndedAtRangeEnd && !connEndedByBackpressure && !connFirstDataSeen`) from the reposition branch so they fall through to the existing ended-connection ladder — status accounting, Retry-After, exponential backoff, `.reconnecting` network phase, bounded give-up. Planned ends (completed range, hard-cap) and any generation that delivered at least one byte keep the fast path, so the #220 no-refetch behaviour is untouched (regression-tested).
- Drop the pinned post-redirect URL (#12) on hard 5xx (503 excluded — rate limiting, #71) and after two consecutive zero-progress failures, so the retry re-resolves through the source URL for a fresh redirect. Expired-per-connection redirect targets (Xtream aggregators) previously pinned forever.
- Log the previously silent non-200/206 response rejection with its status and offset.
- Fire `onVODSourceFailed` when the #169 readError revive cap is exhausted, instead of returning silently with AVPlayer parked in `waitingToPlay` forever. The host-facing message is made mid-session-neutral.
- Test seam: `reconnectBackoffScaleForTesting` (captured at reader init, precedent: `sourceThrottleKbpsForTesting`) so the bounded give-up is testable without a minute of real sleeping.
- `ThrottledOriginServer` gains an injectable per-request directive (`serve206 | status | redirect | dropConnection`) and a path-aware request log; defaults preserve every existing test.

**Alternative considered:** excluding all `connEnded && !connEndedAtRangeEnd` regardless of bytes delivered — rejected as the larger behaviour change; a generation that delivered data keeps today's semantics.

**Test plan**

- `swift test` on macOS 26: 1513 tests / 227 suites green, including three new suites:
  - `ErrorEndedReconnectTests` — spiral bounded (≤20 attempts vs. hundreds) with backoff, `.reconnecting` emitted, `-1` after the cap; a single dropped refill heals with no streak; healthy boundaries stay off the ladder with no `.reconnecting`.
  - `ResolvedURLInvalidationTests` — 302 source → CDN; hard 500 from the pinned CDN URL falls back to the source URL and completes; classifier unit test.
  - `ReadErrorReviveCapTests` — exhausted gate fires `onVODSourceFailed`.
- `aetherctl play` (macOS 26) against a loopback portal (302, expiring tokens) + CDN (500 on suffix ranges, 500 on first attempt at the 32 MiB frontier): heals invisibly — re-resolves through the portal, re-pins, plays to the end; with permanent 500s: bounded ladder → give-up → #169 revive → terminal `onVODSourceFailed` instead of an endless spiral.
- Field shape: Apple TV 4K, tvOS 26, 3.7 GB H.264/AAC MKV via Xtream aggregator (`max_connections=1`, 302→http, no suffix ranges).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,24 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Fixed
+
+- **An origin refusing every range refill no longer spins the reader in an unbounded
+  reconnect spiral.** A connection that ended in error without delivering a byte of its
+  generation was treated as a benign reposition: `seekReconnect` cleared the unproductive
+  streak and applied no backoff, so a connection-capped origin answering 500 at a 32 MiB
+  range boundary produced ~15 reconnects/s (925 in 60 s observed) until the segment
+  provider tore the demuxer down from outside. Such connections now take the failure
+  ladder: status accounting, Retry-After, exponential backoff, a `.reconnecting` network
+  phase, and a bounded give-up. The previously silent non-200/206 response rejection is
+  now logged with its status and offset.
+- **A pinned post-redirect URL is dropped on hard 5xx answers and on repeated zero-byte
+  failures.** Redirect targets that expire per connection (Xtream-style aggregators)
+  answered every later range with 500 from the pinned URL; only auth-expiry statuses
+  (401/403/404/410) invalidated the pin. The reader now falls back to the source URL for
+  a fresh redirect (503 keeps the pin — that is rate limiting, #71).
+- **A VOD session whose readError revive cap is exhausted surfaces `onVODSourceFailed`**
+  instead of dying silently with AVPlayer parked in `waitingToPlay` forever (#169).
 
 ## [6.7.0] - 2026-08-04
 
--- a/Sources/AetherEngine/AetherEngine+Loading.swift
+++ b/Sources/AetherEngine/AetherEngine+Loading.swift
@@ -578,8 +578,9 @@ extension AetherEngine {
                 self.liveSourceReset.send()
             }
         }
-        // #126: zero-progress VOD pump death (readError before any packet/segment). Without this
-        // the host sees isPlayable=true, tracks=0, waitingToPlay until its own first-frame timeout.
+        // #126: zero-progress VOD pump death (readError before any packet/segment), and #169:
+        // mid-session readError after the revive cap. Without this the host sees
+        // isPlayable=true / a stalled item and waits until its own timeout.
         session.onVODSourceFailed = { [weak self, weak session] code in
             Task { @MainActor in
                 guard let self, let session, self.nativeVideoSession === session else {
@@ -589,7 +590,7 @@ extension AetherEngine {
                     )
                     return
                 }
-                self.state = .error("Source read failed before any media was produced (code \(code))")
+                self.state = .error("Source read failed (code \(code))")
             }
         }
         // prepareNativeSubtitles + non-bitmap text tracks: builds the native subtitle table; must be set before start().
--- a/Sources/AetherEngine/AetherEngine.swift
+++ b/Sources/AetherEngine/AetherEngine.swift
@@ -884,6 +884,11 @@ public final class AetherEngine: ObservableObject {
         sourceThrottleKbpsForTesting = max(0, kbps)
     }
 
+    /// TEST-ONLY: scales the reader's reconnect backoff (1.0 = real timing). Read once by each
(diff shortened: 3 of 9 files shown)


superuser404notfound:
Merged, thank you. All three defects hold up under a revert check: with the reposition guard taken back out, `an origin refusing every refill is retried with backoff and given up on` runs past its two-minute limit instead of failing, because the read never returns at all, which is exactly the field symptom; and with the pin invalidation taken out, the retry never reaches the source URL. Full suite green on top of 6.10.0.

Two things pushed on top in 519ae26e:

* The pin is no longer dropped on a rate-limit streak. That clause contradicted the PR's own 503 rationale, and re-resolving spends a second request on an origin that is already metering us, which on a `max_connections=1` panel is the request there is no room for. Hard 5xx and the zero-progress streak are unchanged, and the metered case has its own test now.
* `dropConnection` closed an fd that stays in `_connFDs`, and `stop()` closes those too. Between the two calls the process can hand that descriptor number to a new socket, so the test origin could shut down an unrelated connection. Deregister first, close once.

The `ThrottledOriginServer` directive seam is a good addition, the failure-path tests in this area needed it.

Issue #295, opened 2026-08-03
Software-path SD MPEG-2 (PAL 25i): intermittent brief picture glitches with no engine-log signature — first seen on 6.5.3, unchanged on 6.5.6


rrgomes:
On a 1979 SD television episode (MPEG-2, PAL source, 25 fps interlaced → 50 fields/s after deinterlace, software decode; ~4.5 GB file, local PMS), playback shows several brief visible glitches/pauses per couple of minutes. What makes it report-worthy is the silence around it: our 4 s stall detector never fires, there are no stall/nudge/error lines, DisplayCriteria is correct (content 25.000 → panel 50.000 Hz), and the transport is idle (the pump opens one 32 MB window per ~30 s; no starvation, no reconnects). Renderer cadence holds a steady ~50 enqueues/s throughout; `isReadyForMoreMediaData=false` is pinned from enqueue #8 onward, which may just be a healthily-full pipeline — if so, we have no signal at all coinciding with the visible glitches.

The asset has been part of our regular per-engine gate pass for months and was clean through the 6.3/6.4.0-era engines (we skipped the pass for 6.4.4, so the window is between our last clean pass and 6.5.3; 6.5.4's pump QoS step-down didn't change it, so #286 isn't the mechanism). AppleTV14,1, tvOS 26.5, native output.

Happy to run any diagnostics you'd like enabled, or test intermediate revisions. We see the software path's enqueued-frame count just merged on main (#288) — if a presented/dropped split, or presentation timestamps on the software path (the #260 analogue), is feasible, that would let us catch these moments in the act. We also noticed the post-6.5.6 SAR/decoder work (#290) touches container-declared-PAR handling for exactly this content class (anamorphic SD PAL) — happy to retest on main or the next release if that could interact.

rrgomes:
**The glitch signature is found: every visible stutter coincides with an `[AVIOReader]` detour fill on the direct-URL pump path. The picture defect is IO-timing, not decode.**

New evidence since filing, on 6.6.0 (also reproduced on 6.5.3/6.5.6):

- The glitches are **deterministic by content position** — same places on every linear replay.
- They manifest as a **stutter/jump**, not pixel corruption.
[...]

Issue #174, opened 2026-07-21
EXC_RESOURCE memory-limit crash: native HLS pipeline keeps reading from origin while muxer is correctly backpressure-stalled


G00380316:
### What happened?

VOD playback (routed through the native/loopback HLS-ingest pipeline — `HLSVideoEngine` + `HLSSegmentProducer` + `HLSLocalServer`) crashes with `EXC_RESOURCE (RESOURCE_TYPE_MEMORY: high watermark memory limit exceeded, limit=3376 MB)` after several minutes of otherwise-normal playback.

The engine's own `memprobe` shows the growth is dominated by **compressed memory**, not resident/active memory:

```
t=511s rss=651MB vmInt=606MB vmExt=42MB vmCmp=1916MB physFP=2536MB mallocBlocks=1459697 mallocMB=2401 avioFetchedMB=542 cacheMB=482 subTracks=102 subActive=false
t=541s rss=558MB vmInt=515MB vmExt=42MB vmCmp=2421MB physFP=2950MB mallocBlocks=1606681 mallocMB=2810 avioFetchedMB=576 cacheMB=514 subTracks=102 subActive=false
```

`vmCmp` climbs ~500MB in 30 seconds while `rss` stays flat (and even drops slightly) — memory that was allocated, went cold, and got compressed rather than freed. `subActive=false`/`subCues=0` throughout, so this is unrelated to #166 (already fixed in 5.15.5, confirmed).

At the moment of the crash (device paused in the debugger on the `EXC_RESOURCE`), I checked every AetherEngine-owned thread's stack:

- **`AetherEngine.HLSSegmentProducer.pump`** (serial): blocked in `HLSSegmentProducer.runPumpLoop()` -> `advanceMuxer(to:)` -> `awaitBackpressureRelease(target:head:context:)` -> `SegmentCache.awaitFetchHighWater(reaching:timeout:)` -> `NSCondition.waitUntilDate:` -> `__psynch_cvwait`. **Correctly backpressured**, waiting for the consumer to catch up before muxing more.
- **`com.aetherengine.hls.accept`** (serial): blocked in `HLSLocalServer.acceptLoop()` -> `__accept`. Idle, waiting for a new loopback connection.
- **`com.aetherengine.hls.work`** (concurrent): blocked in `HLSLocalServer.handleConnection(_:)` -> `readHTTPRequest(_:)` -> `__recvfrom`. Idle, waiting for AVPlayer's next segment request.
[...]

Issue #306, opened 2026-08-05
Software path has no network telemetry: expose read bytes / throughput / buffer runway


kskchaitanya1993:
### What we can't show

Our stats overlay has a Network section — throughput, bytes transferred, buffer depth, dropped frames. On the native and loopback paths we read all of it from `AVPlayerItem.accessLog()`. On the **software** path there is no `AVPlayerItem`, so the whole section is blank, and there is nothing on the public diagnostics surface to put there instead.

Measured on the same build, same test bed, back to back:

```
raw .ts  -> backend=native   (loopback, so an AVPlayer exists)
            downloadBps 5891396   bytesLoaded 2369074
            bufferedMs 4961       droppedFrames 0

VP9      -> backend=software  vDecoder "libavcodec VP9 (SW)"
            codec / decoder / colour all present
            nothing under Network at all
```

Same overlay, same code path in our host — the only difference is which backend the engine dispatched to.

### The request

Somewhere on the public surface, for the software path:

- **bytes read** (cumulative, from the source reader)
- **throughput**, or just the bytes counter if you'd rather we derive the rate ourselves — a monotonic counter is honestly the more useful primitive, since a host can window it however it likes
- **buffer runway in seconds** — how far ahead the reader has pulled, distinct from decoded frames queued
- optionally **frames dropped / late**, to match `droppedFrames`

`softwareHostFramesEnqueued` and `segmentCacheDiskBytes` are the only two diagnostics exposed today, and neither answers "is the network keeping up".

### What we tried first, and why it's wrong

`bufferedPosition` looks like the answer and isn't. On a healthy VP9 session it reads **20 ms** — about one frame at 30 fps — so it is the decoded-frame queue, not network runway. Publishing that under a heading called Network would report a near-stall on a stream that is playing perfectly, so we left the rows blank instead. Blank says "unknown"; 20 ms says something false with confidence.

We also noticed 6.7.0's release note that the software path now reports its read-ahead (#303). If that value already exists internally, this request may be mostly "publish what you already compute" rather than new measurement.

### Why it matters to us

The software path is where our exotic content lives — VP9, AV1 without hardware decode, MPEG-4 Part 2 — and it is also where a user is most likely to be near the edge of what their connection or device can sustain. That is exactly when someone opens the stats overlay, and it is currently the one path that can't tell them whether the problem is the network or the decoder. On the native path we can answer that in a glance.

[...]

PR #317, opened 2026-08-07
Backpressure: end the connection at high water instead of suspending the task

### LINKED ISSUE #310: Long-suspended data tasks holding full receive windows starve every Network.framework flow in the app process on tvOS — close-and-rerequest at high water would avoid the dormant state (field dose-response inside)
**Context.** tvOS 26 (AppleTV14,1 ×2), engine 6.7.0, persistent reader, direct-play VOD over LAN. Follow-on to #220's "suspend is advisory": this is what we now believe the long-suspended state does to the whole process.

**Finding.** During low-bitrate playback we see recurring 10–80 s episodes in which EVERY Network.framework-based flow in the app process goes deaf at once — established WebSockets stop receiving (dying of ETIMEDOUT ~30 s in, their sends unACKed), and no NEW nw flow completes a handshake (URLSession and bare NWConnection fail identically, -1001 at deadline, SYN-ACKs arriving on the wire unprocessed) — while in the same seconds raw BSD sockets from the same process complete TCP handshakes AND full plaintext HTTP exchanges 100% (229/229 in one evening), other processes on the box are unaffected, and the process is idle (4–18% CPU, ~16 threads parked, ~30 fds). On tvOS, TCP for nw flows runs in-process (user-space networking); everything observed is consistent with the process's nw inbound path stalling as a unit — beneath CFNetwork, above the kernel-socket path.

**The dose-response tying it to reader dormancy.** The reader suspends its data task at high water and can then sit dormant for long stretches holding a full unread receive window (a closed TCP window on an idle established flow). Dormant-stretch length is set by media bitrate (≈ window size / rate). Same fleet, same wire, three runs in 30 hours:

| media rate | dormant stretch | episodes |
|---|---|---|
| ~1.2 Mbps h264 1080p | ~100 s | ~11 in 80 min |
| ~14 Mbps HEVC 4K | ~8.5 s | 1 in 54 min |
| ~88 Mbps HEVC 4K | ~1.4 s | 0 in 118 min |

A paused player is the limiting case (indefinite dormancy) and is where we've recorded the worst episode (11 unbroken minutes). The effect switches back on at will by returning to low-bitrate content.

**Why the fix belongs at the reader even though the fault looks like Apple's.** Nothing a session owner does dodges the stall once it starts — it hits fresh sessions and bare NWConnection equally. The escape is not entering the dormant state: the engine already ENDS the connection at the hard cap and re-requests on drain (#220's machinery); at low bitrates, though, suspend() engages first and the flow then sits dormant-established ~100 s per cycle. If the reader closed-and-rerequested at high water instead of suspending (or bounded how long a task may stay suspended before the conn is ended), the pathological state never forms — one extra range request per drain cycle (every ~100 s at 1 Mbps; ~2 ms connects on LAN), and a paused player would hold no dormant flow at all.

Happy to run any instrumented build — we have per-second probe arms (URLSession / NWConnection / raw-BSD) that make episodes and their absence directly measurable.

rrgomes:
Closes #310.

**What this changes.** The persistent reader no longer suspends its data task at `winHighWater`. It ends the connection — the mechanism the `winHardCap` escape hatch already used — and the low-water frontier refill re-requests exactly where delivery stopped, through the same path a completed range uses. Nothing is discarded and nothing is re-fetched; the reader now either has an actively delivering connection or no connection at all, and a stalled or paused consumer holds no flow whatsoever.

**Why.** Per #310: a task left suspended holding an undelivered window keeps a dormant established flow whose closed receive window sits unread for as long as the consumer takes to drain — ~100 s per cycle at 1 Mbps, indefinitely while paused — and on tvOS that dormant state correlates, dose-response by media bitrate, with 10-80 s episodes in which every Network.framework flow in the app process goes deaf at once. And by your own #220 measurement, suspend() was never contractual flow control (911 MB arrived after a suspend) — the hard cap existed precisely because the suspend could fail to take. Ending at high water removes the pathological state instead of bounding it.

**Field verification.** Two Apple TV 4K (3rd gen), tvOS 26, 6.7.0 plus this change, a 71-minute synchronized-playback session on the worst-case title from #310's dose-response table (h264 1080p, ~1 Mbps — the regime that previously produced ~9-11 episodes in 80 minutes): zero episodes, zero WebSocket receive errors, zero connection-setup timeouts. ~110 end-and-rerequest cycles per device (one per ~39 s; pump and subtitle-prefetch readers both cycling); the low-water refill won every cycle — the empty-window backstop never fired, so a boundary never became a stall. Two deliberate 3-minute pauses each held zero connections and resumed cleanly. Observed cost: one extra range request per drain cycle, ~2 ms connects on LAN.

**What rides along** (separable if you'd rather take them differently):

- `winHardCap` and its branch are removed — the window is now bounded by construction at high water plus one delivery's in-flight overshoot, well under the cap and the realloc-doubling peak it had to be sized against.
- The suspend machinery (flag, balance-before-cancel paths, post-suspend delivery accounting) is removed, and the deliberate cancel's completion is no longer logged as a transport error.
- `windowDiagnostics` reports `parked` (backpressure-ended, refill pending) in place of `suspended`/`postSuspendBytes`; the memprobe fragment says `Parked=` instead of `Susp=`/`PostMB=`. Easy to restore the old tuple shape if you have consumers.
- Tests: the #174 suite now pins the new contract (a stalled consumer holds no live connection, the parked flag is set, no refill churn while nothing drains, origin bytes bounded; refill liveness and monotonic frontier after a drain). The #295 test's no-overlap assertion is retightened to strictly increasing request starts, since a cancelled range's undelivered tail is now legitimately re-requested — the same offsets-not-counts principle as your eb6ec3f4 hardening, against which this branch is verified.

The core is the high-water end plus the low-water refill; everything else is negotiable.

### DIFF
--- a/Sources/AetherEngine/AetherEngine+Diagnostics.swift
+++ b/Sources/AetherEngine/AetherEngine+Diagnostics.swift
@@ -127,8 +127,8 @@ extension AetherEngine {
                 }
 
                 // #220: the two readers of a subtitled VOD session, separately attributable.
-                // `ahead` above winHighWater (16 MB) with `susp=0` is backpressure that never
-                // engaged; the pump's own window is the control.
+                // `ahead` far above winHighWater (16 MB) with `parked=0` is backpressure that
+                // never engaged; the pump's own window is the control.
                 //
                 // Both paths, not just software. On a direct-play source the native path runs
                 // the HLS loopback, so `HLSVideoEngine` demuxes from the origin through an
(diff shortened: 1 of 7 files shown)


superuser404notfound:
Merged in 1e45bf7e, thank you. The dose-response table in #310 is what carried this: three bitrates, the dormant stretch predicted from window size over media rate, and the episode count tracking it. That is a much stronger case than the usual "this looks suspicious", and the 71-minute two-device run on the worst-case title closes it from the other side.

Taking the whole thing, not just the core: the hard cap and the suspend machinery going away is the point, since a mechanism that is only sometimes honoured is worse than no mechanism plus an honest reconnect. The `Parked=` rename is fine, `pumpIOWindow` and `ioWindowDiagnostics` are internal and nothing outside the engine reads that tuple.

Two things worth knowing:

* One semantic merge fix folded into the merge commit. `pumpIOWindow` landed on main after this branch was cut (#306 telemetry) and still declared the four-field tuple, so the textually clean merge did not compile. Nothing about the design, just the cost of the branch point.
* Follow-up in b2923b76: bytes that arrive after the end are counted and named once per generation when they pass the high water. Each mechanism here has failed the same way exactly once (blocking left the socket reading, suspend was advisory), so if a transport ignores the cancel too, that should read as a line rather than as a climbing memprobe. Same commit prices the #295 guard again on the reader side, since the origin's byte count cannot tell a cancelled range's undelivered tail from a re-fetch.

Full suite green on main, 1544 tests.

Issue #331, opened 2026-08-07
Live TS: high-water backpressure cycles healthy live connections, and reconnects request an unsatisfiable byte frontier


tschuegy:
### What happened?

Playing a live MPEG-TS stream (Xtream-style panel, `LoadOptions.isLive = true`) through the persistent AVIO reader, I expected one long-lived connection. Instead the reader's own backpressure cycles the connection forever, and against some panels the reconnect can never succeed at all:

1) The 16 MB high-water end (#310) has no live branch, and live connections are open-ended by design - so the high-water end is the only thing that ever terminates a healthy live connection. Each end drains ~8 MB to low water and re-requests "at the frontier", but everything broadcast during the drain is gone: the demuxer rejoins on a corrupt TS packet. Panels serve a ring-buffer join burst at line rate on every (re)connect, so the burst refills the window immediately and each reconnect causes the next one - a steady cycle every ~9.5 MB with one mpegts "Packet corrupt" + h264 decode error per cycle, forever.

2) The reconnect request itself carries `Range: bytes=<frontier>-`, but the frontier is reader bookkeeping, not a server-side byte address. A panel that ignores the offset and serves "from now" masks this; a panel that answers 416 to any offset it cannot satisfy turns every reconnect into an unrecoverable rejection loop: same offset ~1/s (read progress from the draining window keeps resetting the unproductive streak, so it never escalates), re-resolving the pin at streak 2 changes nothing (the offset is the problem, not the pin), and the session starves into CoreMedia -12888.

3) Compounding (1): the 509 a connection-capped panel answers while the replaced connection's slot still lingers server-side is classified as a hard 5xx, so every cycle drops the pinned post-redirect URL and re-resolves through the portal - 500-1500 ms extra gap per cycle, and a second request against exactly the origin that has no room for it (the 519ae26e reasoning, left incomplete for 509).

### Steps to reproduce

1. `LoadOptions` with `isLive = true`; url = a live `.ts` from an Xtream-style panel (302 portal redirect to the stream host, max_connections = 1)
2. `engine.load(url:options:)` and let it play
[...]

PR #332, opened 2026-08-07
fix(io): live-correct the persistent reader - window, reconnect shape, and 509 classification

tschuegy:
## Summary

A live source was stuttered on a steady ~30 s cycle by the reader's own 16 MB high-water end, starved entirely against a panel that answers 416 to frontier offsets, and paid a portal re-resolve for every 509. Three tightly-coupled fixes in `AVIOReader`, one per commit.

Reported in #331 (kept open until the reporter retests on device)

## What changed

- **Live window:** live readers run a 64 MB high water (`streamHighWater` parity), instance-captured like `connStallTimeout`. The panel's join burst is absorbed once, steady state plateaus at burst size with the connection never voluntarily ended, and the end-and-refill survives unchanged as the memory backstop for a "live" source that sustainedly outruns realtime. No suspend, no parked task - the #310 contract (actively delivering flow or no flow) holds verbatim.
- **Live reconnect shape:** live persistent connections always request `bytes=0-` (the shape the join already uses and every origin serves); the append anchors delivered bytes at the window frontier exactly as it always has. VOD bounded ranges and frontier offsets are untouched.
- **509 = metering:** one classifier (`isRateLimitStatus`: 429/503/509) used by the pin-drop exemption, both reconnect ladders, the detour fetch, and the Retry-After parse. A lingering-slot 509 keeps the pinned redirect target (completing 519ae26e) and pays the paced rate-limit streak with its bounded give-up instead of the zero-backoff hard-5xx grind.

## Test plan

- Device / OS: Apple TV 4K, tvOS 26.6 (field verification); macOS 26 for `swift test` (1565 tests green)
- Source media (container / video codec + profile / audio / HDR-DV): live MPEG-TS over http from two Xtream-style panels (302 portal redirect, max_connections=1) - H.264 1080i25 + AAC stereo SDR (software path) and H.264 1080p30 progressive + AAC stereo SDR (native/loopback path)
- Result: panel A (ignores Range offsets): the end-drain-reconnect cycle with one `Packet corrupt` + h264 decode error per ~9.5 MB is gone - one connection, `pumpAheadMB` plateaus at burst size, no periodic 509s, 10+ min clean on device. Panel B (416s nonzero offsets, cleanly completes each response after its ~14 MB ring burst): previously starved into CoreMedia -12888 within a minute; now reconnects cleanly through every burst cycle. New loopback tests pin all three behaviors (`LiveWindowBackpressureTests` + 509 cases in `ResolvedURLInvalidationTests`); each was verified failing on the unfixed base, reproducing the field traces verbatim (16.9 MB park mid-burst; one 4 MB burst then starvation; 13 attempts with 12 portal re-resolves on a permanent 509).

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Public API changes are intentional and documented (none: `AVIOReader` is internal, the new `windowHighWater` init parameter is internal test plumbing like `connStallTimeout`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,23 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Fixed
+
+- **A live source is no longer stuttered on a steady cycle by the reader's own backpressure.**
+  The 16 MB high-water end (#310) had no live branch, and live connections are open-ended by
+  design, so ending at high water was the only thing that ever terminated a healthy live
+  connection. Each end drained ~8 MB to low water and re-requested "at the frontier" — a byte
+  offset that means nothing to a live origin — so everything broadcast during the drain was lost
+  and the demuxer rejoined on a corrupt TS packet. And it never happened once: IPTV panels serve
+  their ring buffer as a join burst at line rate on every (re)connect, so the burst refilled the
+  window immediately and each reconnect caused the next one, forever (a field trace against an
+  Xtream panel cycled every ~9.5 MB with a `Packet corrupt` and an h264 decode error per cycle;
+  the loopback repro accepts 17 MB of a 24 MB burst, parks at 16.9 MB and holds no connection).
+  Live readers now run a 64 MB high water (matching `streamHighWater`, the bound the engine
+  already accepts for the other reader that cannot bound by range request): the join burst is
+  absorbed once, steady state plateaus at burst size with the connection never voluntarily
+  ended, and the end-and-refill survives unchanged as the memory backstop for a "live" source
+  that sustainedly outruns realtime.
 
 ## [6.15.1] - 2026-08-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,16 @@ the public-API contract.
   absorbed once, steady state plateaus at burst size with the connection never voluntarily
   ended, and the end-and-refill survives unchanged as the memory backstop for a "live" source
   that sustainedly outruns realtime.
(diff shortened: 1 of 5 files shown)


superuser404notfound:
Merged as 1187bd71. Thanks for this one, the writeup did most of the review work: three separate defects, each with the field trace that produced it and a loopback test that reproduces it.

Verified here before merging, by disabling each guard in turn rather than trusting the green run:

- live high water back to 16 MB: the burst test parks at 16.9 MB mid-burst with `parked=true` and no connection left, which is the cycle from the report.
- live request shape back to the frontier: 4 MB delivered, then starvation, with the same offset repeating.
- 509 out of the rate-limit class: 13 attempts at the boundary with 12 portal re-resolves, matching the number in the PR body.

Full suite green (1623 tests) on the merge result.

One follow-up pushed in #336, because the live request shape needs one more distinction. `bytes=0-` is right for a ring-buffer panel, which has no byte addresses at all. It is wrong for the other live shape that reaches `AVIOReader` through the same `LoadOptions.isLive` flag: a live source that IS a growing file (a Jellyfin live stream file, or a misdeclared VOD) answers the frontier correctly and resumes there, and asking that one for byte zero re-delivers its whole buffer on top of the window. Since `isLive` is a host declaration, the flag cannot tell the two apart, so the follow-up lets the origin decide: the frontier is kept until a 416 comes back on a nonzero live offset, and that rejection latches the join shape for the rest of the session. Panel type B pays one rejected request per reader instead of one per reconnect, so the log will carry a single `live origin rejected offset N (416); requesting the stream as a join from here` line early in the session and none after it.

#331 stays open until the retest on device.

Issue #176, opened 2026-07-21
DV P5 displaying a green-purple color shift


ijuniorfu:
### What happened?

<img width="1169" height="749" alt="Image" src="https://github.com/user-attachments/assets/08d1da8e-7164-482e-9be9-2f1f90127a5f" />

### AetherEngine version or commit SHA

3e6bdbc

### Host app

AetherPlayer

### Platform

macOS

### OS version

macOS 26.5.1

### Device / chip

Mac mini m4

### Playback path

Software (dav1d) host

### Source media (for playback bugs)

MKV, HEVC, Dolby Vision Profile5, 3840x2160

<img width="1003" height="181" alt="Image" src="https://github.com/user-attachments/assets/81a27cf3-99ce-4753-ae72-fc8ebb953fb5" />

### Error codes / log lines

```text

```

### Anything else

[p5.error.log.txt](https://github.com/user-attachments/files/30226105/p5.error.log.txt)

DrHurt:
P5 cannot be correctly decoded by software path as it has no HDR10/HLG fallback layer like P8.x 

Why are you using the software path with P5?

ijuniorfu:
> P5 cannot be correctly decoded by software path as it has no HDR10/HLG fallback layer like P8.x 
> 
> Why are you using the software path with P5?
> 

I don't know, but this video will be like this. The other p5 is hardware

DrHurt:
So you’re saying you’re not manually forcing software decode then…

Odd. This is a routing bug then! HEVC P5 should always go down the native AVPlayer pipeline.

Let’s see what Vincent has to say. Might have to do with the recent VT probe changes.

superuser404notfound:
Fixed in [5.17.5](https://github.com/superuser404notfound/AetherEngine/releases/tag/5.17.5) (PR #182).

[...]

Issue #133, opened 2026-07-17
Live ingest: SPS/PPS/IDR gate needed for mid-stream H.264 joins (green frames / total muxer failure), plus DisplayCriteria SET/waitForSwitch bypass for unchanged-rate live zaps


digilearn-dev:
### What happened?

Two related but distinct issues found during live TV testing, both on the live HLS ingest path (HLSSegmentProducer, HLSVideoEngine).
Issue A — SPS/PPS/IDR gate missing on live join. When the ingest joins a live MPEG-TS stream mid-broadcast, the existing keyframe gate (video gate open, gated on AV_PKT_FLAG_KEY) opens on any flagged keyframe without requiring valid SPS/PPS to already be parsed, or verifying the keyframe is a true IDR with usable parameter sets. Two severities observed:

Mild: decoder briefly renders with an uninitialized/garbage reference (green frame) before self-correcting once real SPS/PPS/IDR arrive. Seen on multiple channels.
Severe: SPS/PPS loss is complete enough that frame dimensions never resolve at all, allocateMuxer gets 0x0 from probe-time codecpar, avformat_write_header fails, and the channel produces zero segments — an empty #EXTM3U playlist that never recovers without the user manually zapping away.

I expected the engine to withhold packets until a decodable SPS/PPS + IDR sequence is established (the way H264SPS.extractSPSandPPS / extractAdVideoConfig already does for SSAI ad-pod codec switches elsewhere in the same file) — instead the initial live join has no equivalent gate, so both failure modes above can occur.
Issue B — DisplayCriteria RESET defeats the engine's own zap optimization. Custom app's live-zap teardown was calling aetherEngine.stop() with default resetDisplayCriteria: true before loading the next channel. This nils preferredDisplayCriteria, forcing the subsequent load()'s apply() to renegotiate the display mode from scratch even when the outgoing and incoming channel have identical rate/format — and on our panel (unobservable-DV switch signaling) that renegotiation burns the full ~3000ms waitForSwitch() cap every time, dominating zap latency (ttff_ms consistently 2.5–3.0s for load_profile=liveZap).
[...]

Issue #274, opened 2026-08-01
waitForSwitch() still runs when suppressDisplayCriteria=true, adding a residual startup wait even when the host is the sole criteria writer


digilearn-dev:
### What happened?

With suppressDisplayCriteria=true (production mode), Aether correctly skips its own apply() (only runs .clearStale), leaving the host as the sole writer of display criteria. However, the play-gate waitForSwitch() still runs unconditionally on this path, regardless of suppressDisplayCriteria.

This surfaced as a real bug: our host-side rate write, made mid-load once the source frame rate was known, happened to overlap with this unconditional waitForSwitch(). Since Aether itself never applied anything (didApply == false on this path), the wait fell into the wrong branch and mis-logged a real, host-driven SDR rate switch as a failed HDR one:

[DisplayCriteria] WARN switch ended (~2450ms) but EDR headroom still 1.0 (panel stayed SDR despite HDR criteria)

That cost ~2.4s of VOD startup time on affected sessions. We fixed the overlap on our side by deferring the host's rate write until after engine.load() returns, which eliminated the WARN and the mislabeling entirely — confirmed across several fresh VOD plays, zero recurrences.

However, comparing before/after ttff_ms on network-latency-matched sessions, we only recovered part of the lost time (roughly 1.3–1.4s of the original ~2.4s). The likely explanation, and the reason for this report: waitForSwitch() is still called on every load even with suppressDisplayCriteria=true, and even with no host write overlapping it, it appears to still cost a real amount of time (a "Stage 1: no switch started" empty wait) rather than returning immediately. If the host has declared itself the sole criteria writer, we'd expect Aether's own play-gate to either skip this wait entirely or resolve it near-instantly when there's genuinely nothing for Aether to be waiting on.

### Steps to reproduce

<ol dir="ltr" class="[li_&amp;]:mb-0 [li_&amp;]:mt-1 [li_&amp;]:gap-1 [&amp;:not(:last-child)_ul]:pb-1 [&amp;:not(:last-child)_ol]:pb-1 list-decimal flex flex-col gap-1 pl-8 mb-3 print:block print:space-y-1"><li class="font-claude-response-body whitespace-normal break-words pl-2">Configure a session with <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">suppressDisplayCriteria=true</code>.</li>
[...]

PR #338, opened 2026-08-08
Classify a late in-progress flag as an in-gate start, and spend budgets as deadlines (Re Sodalite#49)

superuser404notfound:
Second round of measurement work on Sodalite#49. The reporter's three new logs are the first unfiltered ones, and they settle one question, expose two defects in the instrumentation, and produce one number worth keeping.

## What the logs settle

The panel is innocent for the third time: `content=29.970 → panel nominal 59.943` in all three runs. And the judder is gone on the current build.

## Defect 1: `pre-gate` was reported for a switch that started inside the gate

All three runs came back `start pre-gate`, including the flipped run at `after 376ms`. Stage 1 tested `isDisplayModeSwitchInProgress` on every poll but drew the conclusion `StartSignal.preGate` documents for the first one only ("already set when polling began"). A flag observed false at entry cannot describe a switch that predates entry, so that run was logged as the exact ordering the issue was filed on while being its opposite.

`classifyStart` is now a pure function with a third case, `inGateFlagOnly`, for a flag that rises later with no start notification. The switch began inside the gate either way; the missing piece is the notification, which points at the observer registration point (they are registered on gate entry) rather than at the panel.

## Defect 2: budgets were poll counts, not deadlines

`n` sleeps of `m` ms is only `n * m` on an idle scheduler. The reporter's Apple TV reports `thermal=serious` from the first sample onward in all three logs, and the same 40 x 50 ms Stage 2 took **2082 ms** in one run and **2862 ms** in another. Both stages now measure elapsed time against `budgetMs` / `stage2CapMs`.

This is the second measurement defect in a row on this path; the first one (#279) inflated every settle time by the whole Stage 1 budget.

## The number worth keeping

The flipped run is the only one where the panel ever reported an end, and it came at **3238 ms total** (2862 ms after the start was seen). Two of three runs instead hit the Stage 2 cap with no end reported at all, both rate-only SDR. So on this panel a rate-only switch is neither sub-second nor reliably observable, which contradicts the d87b54d premise `.brief` rests on.

**Not changed here:** the cap is not raised and `.brief` is not widened. One panel is not grounds for making every session's cold start wait longer, and the runs that produced these numbers played cleanly. The measurements are in the doc comments so the next change to those budgets starts from evidence rather than from a premise.

## Also not changed, deliberately

Registering the switch observers at the criteria write instead of at gate entry would be the real fix for "end never reported", and `inGateFlagOnly` is precisely the evidence that would justify it. It touches `apply()`, the path carrying DV P5 cold start, #133 and #176, and it cannot be verified without hardware. The order that was missing last time: be able to measure first, then rebuild. If the next round shows `inGateFlagOnly` dominating, the rebuild has a target and a way to tell whether it worked.

## Tests

23 in the suite, 7 new, covering the classification table and the deadline behaviour with the reporter's two measured cap durations as cases. Full run: 1631 tests in 241 suites green, plus a `generic/platform=tvOS` build, since the changed loops sit under `#if os(tvOS)` and `swift test` on macOS never compiles them.

https://claude.ai/code/session_01HX5zV3Fcf7Nzq4DYGdXvQE

### DIFF
--- a/Sources/AetherEngine/Display/DisplayCriteriaController.swift
+++ b/Sources/AetherEngine/Display/DisplayCriteriaController.swift
@@ -70,27 +70,44 @@ final class DisplayCriteriaController {
         /// Nothing can be pending; return before touching the display manager.
         case skip
         /// 200 ms: only a rate-only switch could still start (the pre-#274 budget, sized for a synchronous
-        /// engine write). The engine already declines to pre-flight its own rate-only writes on the premise
-        /// that they are sub-second, which is an assumption from d87b54d that no log has ever measured: the
-        /// settle times it would have been read off were inflated by the whole Stage 1 budget (Sodalite#49).
-        /// `timingSuffix` now reports the real number, so this budget can be confirmed or corrected.
+        /// engine write). The engine declines to pre-flight its own rate-only writes on the d87b54d premise
+        /// that they are sub-second.
+        ///
+        /// That premise is now contradicted, though not yet by a run that also juddered: on the Sodalite#49
+        /// reporter's Apple TV a rate-only SDR switch was measured at ~2.9 s end to end, and two further
+        /// runs never reported an end at all inside Stage 2's cap. Reading that as "shorten nothing, widen
+        /// nothing" is deliberate: one panel is not grounds for making every session's cold start wait
+        /// longer, and the runs that produced these numbers played cleanly.
         case brief
         /// 1000 ms: a dynamic-range switch may still be inbound from a writer whose timing we don't control
         /// (AVKit's auto-criteria path fires from the AVPlayerItem formatDescription). DV P5 cold start
         /// depends on this budget.
         case full
 
-        var ticks: Int {
+        /// The budget both as the log lines report it and as Stage 1 spends it. Latency fixes are verified
+        /// off the number they moved, so both the spent-time line and the #289 skip line name it (#274).
+        var budgetMs: Int {
             switch self {
             case .skip:  0
-            case .brief: 20
-            case .full:  100
+            case .brief: 200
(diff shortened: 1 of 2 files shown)


superuser404notfound:
## One behavioural consequence worth deciding before this merges

The deadline change is not purely cosmetic for Stage 1, and I understated that in the description.

`.full` was 100 iterations of `Task.sleep(10ms)`. On an idle scheduler that is the documented 1000 ms. Under the load these logs were captured at, it is not: the same drift that stretched Stage 2's 40 x 50 ms to 2862 ms applies here too, and short sleeps carry proportionally more overhead than long ones. So the effective Stage 1 wait under load has probably been somewhere north of 1000 ms, and this change makes it exactly 1000 ms.

That is a shortening, and it lands on the one guarantee this gate exists for: DV Profile 5 cold start, where starting the decode before the switch begins races an AVPlayer error. #274 verified the budget against the log line that printed `budgetMs`, i.e. against the nominal number, so a longer real wait would not have shown up in that verification either.

Three ways to read it:

1. **Correct as written.** 1000 ms is the documented, reviewed and #274-verified value. If the real wait was longer, that was an accident, and restoring the intended value is the fix rather than a regression.
2. **Deadline for Stage 2 only**, leaving Stage 1 counting ticks. Removes the risk, keeps the inconsistency, and leaves Stage 1's reported budget still not matching its real one.
3. **Deadline for both, with `.full` re-sized** to whatever the effective wait has actually been. No number to base that on today.

I have left it at 1, because the alternative is keeping a budget that is only correct when nothing else is happening. But it wants a DV P5 cold start on hardware before it ships, not just green CI, and that is not something the test suite can cover.

Issue #199, opened 2026-07-22
Rerouted live-ingest session death relands on the known-bad native path: #168 verdict not remembered, no in-engine revive


superuser404notfound:
Split out of #189 (deep-window follow-up by @kskchaitanya1993). The #189 holdback fix itself is verified there (loopback `-16832` = 0 across a 2.5 min run with the live window deepened to ~22s, comfortably past the ~18s holdback runway). This issue tracks what remains: the session still cycles through the `#168` reroute dance roughly every 13s instead of settling into stable play.

## Observed (5.18.5-family, Apple TV 4K 3rd gen, tvOS 26.5)

HEVC-in-MPEG-TS live channel, test-bed loops a fixed segment pool and injects a real `[DISCONTINUITY]` each lap. Over 2.5 min: 8 `#168` reroutes, 0 host ladder re-tunes, playback never leaves the engine, it just loops:

```
#9  load 192.168.x/hevc-hdr/master.m3u8   -> readyToPlay -> playing (native, audio-only)
#9  master advertises video but AVPlayer built no video track after grace
#168 reroute -> #10 load 127.0.0.1/media.m3u8 (loopback)
     live seg-0 ... seg-1 [DISCONTINUITY] ... seg-2 ... seg-3
#11 load 192.168.x/hevc-hdr/master.m3u8   <- origin master again, native bypass
#11 built no video track -> #168 reroute -> #12 loopback -> repeat
```

## Root cause: two engine-side defects composing into a loop

**Defect A: the `#168` verdict is not remembered.** The video-carriage watchdog (`RemoteHLSIngestFallback.Watchdog`) discovers "this master advertises video but AVPlayer will never build a video track for its MPEG-TS carriage" at runtime, at the cost of a full native mount, readyToPlay, and the 8-tick (~4s) grace. The verdict is then thrown away. Every subsequent `load()` of the same URL with `nativeRemoteHLS` re-mounts the doomed native bypass and re-pays the whole discovery cost. Nothing in the engine records that this master is a known reroute case.

[...]

PR #341, opened 2026-08-08
fix(video): give live muxerFailed pump exits a recovery arm

tschuegy:
## Problem

A **live** pump death with `muxerFailed` has no recovery arm: `handlePumpFinished` hits a bare `return`, `shouldHaltLiveProduction` returns `false` for it (so the provider is not even halted), and no host signal fires. The provider keeps serving a frozen playlist, AVPlayer parks on it waiting for buffer that never comes, and the session zombifies until the viewer zaps away.

Field trace (immediately after the rotation death that #340 fixes at its root):

```
[HLSSegmentProducer] pump finished: reason=muxerFailed packetsRead=39549 packetsWritten=24320 lastError=0
[NativeAVPlayerHost] #7 errorLog ... -12888 - Playlist File unchanged for longer than 1.5 * target duration  (repeats forever)
[LagDiag] clk=483.66 dclk=0.00 tcs=waiting rate=1.00 wait=AVPlayerWaitingToMinimizeStallsReason fwd=2.0 ... (frozen until app exit)
```

The `shouldHaltLiveProduction` comment assumed "AE#222 rebuilds into the same provider" — that was doubly false for live:

1. `muxerFailed` never reaches the AE#222 arm at all (bridged sessions are excluded from the exit-scan, so the reason is never converted).
2. Even the AE#222 arm itself was a silent no-op for live: its live paths returned without doing anything, and the admitted rebuild goes through `requestRestart`, which bails on the empty live `segmentPlan` (`performRestart`'s `idx < segmentPlan.count` guard) — a second zombie path, hit by non-bridged E-AC-3 rotations.

## Fix

New `handleLiveMuxerFailure`: rebuilds the producer **in place** on the same demuxer and provider at `liveContinuationPoint()` — `finishLiveReopen` minus the fresh connection. A reopen is deliberately not used: the socket is healthy (the muxer died, not the connection), and `performLiveReopen` opens its new connection *before* closing the old demuxer, which would double-connect against `max_connections=1` portals.

- Bounded by **progress**, not per session: new segments since the last death reset the budget (an hours-long channel legitimately crosses several encoder restarts); three consecutive barren deaths exhaust it.
- On exhaustion (or a producer-build throw): `markLiveProductionHalted()` — releasing held blocking-reload waiters — and `onLiveSourceReset` so the host can retune. Same last-resort surface as #199's reopen exhaustion.
- Live `.needsAudioSampleEntryPrime` is routed through the same in-place rebuild (the captured prime is already in `sessionAudioMoovPrimeFrame`), closing zombie path 2.
- `shouldHaltLiveProduction` still returns `false` for both reasons — an eager halt would 503 the very provider the rebuild serves — but the comment now names the owning arm, and both arms halt themselves on budget exhaustion.

## Tests

- `liveMuxerRebuildDecision` extracted as a pure decision and pinned in `LiveProductionHaltTests`: first death always rebuilds; progress resets the budget; exactly the cap's worth of barren rebuilds, then halt + host retune.
- Existing halt-classification assertions updated with rationale strings naming the new arm.

Root-cause companion: #340 prevents this particular muxer death entirely; this arm is what keeps *any* future live muxer death from zombifying the session.

### DIFF
--- a/Sources/AetherEngine/Video/HLSVideoEngine+LiveReopen.swift
+++ b/Sources/AetherEngine/Video/HLSVideoEngine+LiveReopen.swift
@@ -72,7 +72,9 @@ extension HLSVideoEngine {
         case .eof, .readError, .keyframeStarvation:
             return !sourceReopenable
         case .stopRequested, .muxerFailed, .backpressureWedge, .needsAudioSampleEntryPrime:
-            // AE#222 rebuilds into the same provider with a primed muxer, so production continues.
+            // muxerFailed / needsAudioSampleEntryPrime: handleLiveMuxerFailure and the AE#222 arm rebuild
+            // the producer into the SAME provider, so an eager halt here would 503 the very provider the
+            // rebuild is about to serve; both arms halt themselves once their budget is exhausted.
             return false
         }
     }
@@ -95,6 +97,21 @@ extension HLSVideoEngine {
         return .none
     }
 
+    /// Live muxerFailed pure decision: the muxer died but the connection is healthy, so the arm rebuilds
+    /// the producer in place — bounded by PROGRESS, not per session: a rebuild that cut new segments since
+    /// the last death earns a fresh budget (an hours-long channel legitimately crosses several encoder
+    /// restarts), while `cap` consecutive barren deaths mean the source is not muxable as-is and the
+    /// session must halt production and delegate to host retune.
+    static func liveMuxerRebuildDecision(
+        continuationIndex: Int,
+        lastContinuationIndex: Int,
+        barrenCycles: Int,
+        cap: Int
+    ) -> (rebuild: Bool, newBarrenCycles: Int) {
+        let cycles = continuationIndex == lastContinuationIndex ? barrenCycles + 1 : 0
+        return (cycles < cap, cycles)
+    }
+
     func handlePumpFinished(_ prod: HLSSegmentProducer,
                                     reason: HLSSegmentProducer.PumpExitReason) {
         // #65 (VOD only): a broken backpressure wedge means AVPlayer is stuck behind a parked producer.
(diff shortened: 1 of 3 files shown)


superuser404notfound:
Merged. Verified onto current main (the branch was based on 6.15.2): 1689 swift-testing plus 468 XCTest green, tvOS Simulator build clean, and the cap test is load-bearing (flipping `cycles < cap` to `<=` fails it on 4 vs 3).

The two diagnoses check out against the source, both of them:

- `performRestart` really does bail on live at `idx < segmentPlan.count` with an empty plan, so the admitted AE#222 rebuild consumed its gate and rebuilt nothing. Silent, because the guard returns before it touches anything.
- The in-place rebuild is the right tool rather than a reopen, and `liveReopenOutputEndSeconds` is what makes it safe: it sets `videoTarget = Int64.min` and stamps the first tfdt at the continuation point, so wiring `onVideoShiftKnown` to the rebase handler is required, not incidental. `stop()` nils the producer under `restartLock` before it joins the pump, so the `producer === failed` guard covers the teardown race and there is no lock inversion from the pump thread.

One thing pulled in on top in 513fd839: the reopen ladder directly above `handlePumpFinished`'s new arm has always made the identical decision from an inline copy (continuation point vs the last one, count barren attempts, give up at 3), untested. Both arms now call the shared `liveRecoveryBudgetDecision`, renamed off the muxer arm since it is no longer only about muxer rebuilds. Same predicate over the same counters, each arm keeps its own counter pair.

Also written down at the decision rather than in either arm: a source that dies after every handful of segments never goes barren and recovers forever. Deliberate, since bounding it needs a rate rather than a counter and an eager cap would kill the long-running channels the progress budget exists for, but worth naming where the next reader will look.

Ships in the next release, together with #340.

Issue #93, opened 2026-06-30
Loopback-HLS: a single backward seek on heavy 4K can wedge the segment producer (video stalls, audio continues -> A/V desync)


rrgomes:
**Symptom.** On heavy 4K content over the loopback-HLS path, a **single backward seek (not a rapid burst)** can intermittently wedge playback: the **video freezes while AVPlayer's already-buffered audio keeps playing**, producing A/V desync, periodic audio dropouts, and dropped subtitles. The producer logs a sustained backpressure park and the re-anchor cap is hit:

```
[HLSSegmentProducer] #65 backpressure PARK (advance) ... parked=Ns (no playback progress)
[HLSSegmentProducer] #65 backpressure WEDGE BROKEN (advance) ... exiting pump for host re-anchor on AVPlayer position
[HLSVideoEngine] #65 backpressure wedge re-anchor cap reached (6 consecutive at pos=...); giving up (AVPlayer not resuming)
[NativeAVPlayerHost] item.status=failed err=AVFoundationErrorDomain/-11819 'Cannot Complete Action'   (in the worst case)
```

Around the same area the mp4 muxer logs malformed packet timing:

```
[mp4] [warning] Packet duration: -656 / dts: 23840480 in stream 0 is out of range
```

**Intermittent.** In the same session, earlier seeks resolved cleanly -- `frameAhead / producerShift / hostShift = 0.000` throughout, the seek landed and resumed normally -- while a later, comparable moderate seek wedged. So it's timing/state-dependent, not deterministic.

**Asset.** 4K (3840x1634) HEVC, DV Profile 8.1, E-AC3 5.1, MKV, ~25 Mbps. The wedge is on the backpressure/loopback path, so it likely is **not DV-specific** -- heavy 4K is the common factor.

This is in the #65 family you've fixed iteratively (4.2.2 / 4.5.2 / 4.5.3 / 4.5.4). On heavy 4K it still bites occasionally on a single seek. Logs and asset available; happy to test patches.

superuser404notfound:
Thanks for the detailed report.

Since you filed this, we shipped a fix for #92 that reworked VOD segment cutting (keyframe-gated, so the IRAP opens its own segment). It touched a related but distinct code path, so before I patch #93 I'd like to know whether it still reproduces on the current build.

[...]

Issue #65, opened 2026-06-24
Presented frame drifts ~6 s ahead of currentTime/sourceTime after a rapid bidirectional seek burst (loopback-HLS native path) — frameAhead/producerShift/hostShift all report 0


rrgomes:
Filing as the Watch Together consumer that just adopted `isSeeking`/`seekTarget` (#42/#38 — working well for us, thanks). Seen on **3.13.4, tvOS 26**, native loopback-HLS direct play (H.264 title with a PGS subtitle track).

**Symptom.** After a rapid *bidirectional* seek/scrub burst, the engine settles into a state where the frame on screen is **~6 s ahead, in source content, of what `currentTime` and `clock.sourceTime` report** — and it stays there (persists through a pause; does not reconcile). Hosts read `currentTime` for the scrubber and `sourceTime` for the subtitle clock, so the timeline and subtitles both trail the picture by ~6 s.

**Frame-exact verified:** with playback paused, opening the same file in IINA and stepping to the exact frame on screen read ~6 s later than the engine's reported position (not an eyeball estimate).

**The engine's coherence metrics did not reflect it.** At that moment, reading the public accessors:

```
currentTime=181.50  sourceTime≈181.5  avpClock=181.58  frameAhead=0.000  producerShift=0.000  hostShift=0.000
```

`currentTime ≈ avpClock`, and `frameAhead = producerShift − hostShift = 0` — so the engine believed it was fully coherent while the presented content was ~6 s further along. (`avBufAhead` was ~6.5 s per the memprobe — suggestive, but I won't over-read it.)

**Trigger / repro.** Hit via a two-peer "dueling scrub" — both peers scrubbing opposite directions, ~10 `seek(to:)` on one engine within a few seconds. The seek *pattern* (rapid **alternating-direction** seeks on the loopback-HLS native path) should reproduce solo — like the #49 seektest harness but alternating direction rather than monotonic. Looks like a sibling of #49 (closed) where the divergence lands between the *presented source content* and `currentTime`, with the shift accounting reading 0.

**Hypothesis (low confidence — you know the internals):** the producer's source-PTS → playlist-time mapping drifts by ~6 s under the alternating-direction burst, but the shift used to derive `currentTime`/`sourceTime` settles back to 0 without capturing the offset — so AVPlayer faithfully presents the playlist position whose bytes are ~6 s of source ahead of the reported clock.

[...]

PR #342, opened 2026-08-08
fix(playback): make the stall watchdog level-triggered with a host escalation rung

tschuegy:
## Problem

The #65 stall re-engage watchdog is one-shot and edge-triggered: armed per `playbackStalled` notification, then a **single instantaneous check** 6 s later. Any media fetch inside that grace window disarms it permanently — only a new `playbackStalled` re-arms.

Field trace: a live playlist froze at its last segment (pump death; recovery arm added in #341). AVPlayer stalled, drained the remaining tail segments (fetches inside the grace window → watchdog dead), then parked in `waitingToMinimizeStalls` with ~2 s buffered:

```
[NativeAVPlayerHost] #7 playbackStalled
[LagDiag] clk=483.66 dclk=0.00 tcs=waiting rate=1.00 wait=AVPlayerWaitingToMinimizeStallsReason fwd=2.0 keepUp=n empty=n ... (forever)
```

In that parked state nothing can ever fire again: `playbackStalled` needs a buffer-exhaustion transition (buffer stays non-empty), a waiting player never posts `failedToPlayToEndTime`, and the producer-side wedge detector died with the pump. The freeze was terminal from the couch.

## Fix

Three changes; together they close every exit from the trap:

1. **Level re-watch.** Fetch activity during the grace window re-baselines and keeps watching (bounded by `maxStallWatchPasses = 10` ≈ 60 s) instead of returning permanently. Only a fully silent window escalates into the unchanged nudge → stage-2-reload ladder; a recovered/paused/failed player still disarms (those states have other owners). Verdict extracted as a pure function.

2. **Persistent reload budget (storm shape).** On a frozen playlist each stage-2 reload replays the tail and re-stalls within seconds, and the fresh stall *supersedes* the ladder task before any post-reload check can run — a per-task rung alone can never fire. A gate spanning stall events (same shape and epsilon as the #93 item-death gate) admits two reloads at the same frozen position; after that the ladder publishes `liveSourceReset` instead of reloading again. Real progress restores the budget.

3. **Post-reload rung (silent-park shape).** When no further stall supersedes, a rendered clock that has not moved a whole window (12 s) after the stage-2 reload while the player still waits publishes `liveSourceReset`: the local session is unrecoverable consumer-side and only the host can retune. Live-gated; decision extracted as a pure function.

`liveSourceReset` is the existing host contract ("negotiate a fresh URL and call load"); this only adds the two missing publishers for consumer-side dead ends.

## Tests

`StallWatchdogLevelRearmTests` pins the incident case (fetch-during-grace must re-watch, not disarm), escalation on silence, disarm on recovery/failure/cap, the frozen-clock rung (live-only), and the storm-shape budget (two reloads at a frozen position, then escalate; progress restores).

### DIFF
--- a/Sources/AetherEngine/AetherEngine+Loading.swift
+++ b/Sources/AetherEngine/AetherEngine+Loading.swift
@@ -1038,17 +1038,43 @@ extension AetherEngine {
                 let fetchesAtStall = self.nativeVideoSession?.mediaFetchCountSnapshot ?? 0
                 self.stallReengageTask?.cancel()
                 self.stallReengageTask = Task { @MainActor [weak self, weak host] in
+                    // Level re-watch (#65): fetch activity inside the grace window used to disarm
+                    // this watchdog permanently — but a player that drains its remaining TAIL
+                    // segments and then parks on a frozen playlist (fwd buffer non-empty, so
+                    // playbackStalled never re-fires) was exactly that case, and nothing ever
+                    // re-armed. Re-baseline and keep watching instead, bounded so trickling
+                    // fetches on a merely slow session hand back to the producer-side arms.
+                    var baseline = fetchesAtStall
+                    var passes = 0
+                    watch: while true {
+                        try? await Task.sleep(
+                            nanoseconds: UInt64(Self.stallReengageGraceSeconds * 1_000_000_000))
+                        guard !Task.isCancelled, let self, let host,
+                              host.stallCount == count,
+                              let player = self.currentAVPlayer else { return }
+                        let fetchesNow = self.nativeVideoSession?.mediaFetchCountSnapshot ?? 0
+                        switch Self.stallWatchVerdict(
+                            fetchesNow: fetchesNow,
+                            baseline: baseline,
+                            isWaitingToPlay:
+                                player.timeControlStatus == .waitingToPlayAtSpecifiedRate,
+                            itemFailed: player.currentItem?.status == .failed,
+                            passesSoFar: passes,
+                            cap: Self.maxStallWatchPasses
+                        ) {
+                        case .disarm:
+                            return
+                        case .escalate:
+                            break watch
+                        case .rewatch:
+                            passes += 1
+                            baseline = fetchesNow
+                        }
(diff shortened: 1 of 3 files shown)


superuser404notfound:
Merged. Verified onto current main (the branch was based on 6.15.2): 1689 swift-testing plus 477 XCTest green, tvOS Simulator build clean, and the watch cap is load-bearing (flipping `passesSoFar + 1 < cap` to `<=` fails the trickle test).

The trap analysis holds all the way down. `mediaFetchCount` increments in `handleTargetChange`, so it counts request ARRIVALS at the local server: a player draining tail segments really does bump it inside the grace window, and the old single check really did return for good. `renderedTime` comes off a 100 ms periodic observer, and `reloadStalledConsumerItem` reloads with `inPlaceSwap: true`, which returns before the `renderedTime = 0` reset, so the post-reload baseline survives the swap.

One change on top in b4956c18: the final rung asked whether the clock CHANGED, while its own doc comment promised whether it moved. Since stage 2 is an in-place swap, a swap that half took parks the fresh item on its own clock (or on zero before it is ready) and is exactly as dead, but equality reads that as recovery, so the rung would sit out the shape it exists for. It now asks whether the clock advanced past the same 0.5 s progress epsilon `ItemDeathReviveGate` and `shouldSurfaceDeferredFailure` already use. Widening it is safe because the clock was never the false-positive guard, `isWaitingToPlay` is, and nothing reaches the rung without a stall, two silent grace windows and a reload in front of it.

Also split the two rung tests so each moves one variable (the old moved-clock case flipped the clock and the player state together, so it pinned neither) and added the parked-on-a-different-clock case.

Worth naming for the retest, since it is inherited rather than new: the escalation still triggers on a full grace window with zero segment REQUESTS while waiting, and the level re-watch gives that condition 60 s of chances instead of 6. A link slow enough to spend more than a window on a single segment can therefore reach the nudge where it previously would not. Two silent windows are still required for the reload, so the exposure is the same shape as before, just observed for longer.

Ships in the next release with #340 and #341.

PR #343, opened 2026-08-08
fix(video): escalate live reopen exhaustion for URL sources too

tschuegy:
## Problem

Both live reopen-exhaustion sites escalate only for the `.customFactory` transport that #199 introduced:

- the barren-cycle cap in `handlePumpFinished` ("live source produced no segments across N reopen cycles; giving up"), and
- the attempt cap at the end of `performLiveReopen` ("live reopen FAILED after N attempts; source considered permanently lost").

A **URL-source** live session that exhausts its reopen budget reaches the same terminal state with neither `markLiveProductionHalted()` nor `onLiveSourceReset`: the provider keeps advertising blocking reloads it can never satisfy (the -15410 shape the halt latch exists for), the playlist stays frozen, and the host is never asked to retune. That is exactly the zombie window #199 closed for engine-created ingest readers — still open on the most common transport.

Found while tracing a field freeze (same session as #340/#341/#342): the incident itself died through the live-`muxerFailed` hole #341 covers, but this adjacent exhaustion path has the identical failure shape for URL sources whose reopens run dry.

## Fix

Extract the escalation decision as a pure function, `liveReopenExhaustionEscalatesToHost(transport:)`, covering **every** reopenable transport, and use it at both sites. `.none` deliberately stays out: a transport-less source delegates to host retune before any reopen begins, and escalating it again would double-signal `liveSourceReset`.

Behavior for `.customFactory` is unchanged; `.url` now halts + signals on exhaustion instead of zombifying.

## Tests

Pinned in `Issue199RerouteRecoveryTests` next to the existing transport-decision pins: `.url` and `.customFactory` escalate on exhaustion; `.none` never reaches the escalation.

### DIFF
--- a/Sources/AetherEngine/Video/HLSVideoEngine+LiveReopen.swift
+++ b/Sources/AetherEngine/Video/HLSVideoEngine+LiveReopen.swift
@@ -95,6 +95,17 @@ extension HLSVideoEngine {
         return .none
     }
 
+    /// #199 follow-up pure decision: EVERY in-engine reopen transport that exhausts its budget must
+    /// halt production (dropping the blocking-reload advert, releasing held ?_HLS_msn= waiters) and
+    /// surface the loss to the host via onLiveSourceReset. #199 scoped that escalation to the factory
+    /// transport it introduced, which left a URL source's exhaustion as a zombie session: provider
+    /// un-halted, playlist frozen, host never told to retune. `.none` never reaches an exhaustion
+    /// site — it delegates to host retune before any reopen begins — and must stay out so a future
+    /// call-site reshuffle cannot double-signal that path.
+    static func liveReopenExhaustionEscalatesToHost(transport: LiveReopenTransport) -> Bool {
+        transport != .none
+    }
+
     func handlePumpFinished(_ prod: HLSSegmentProducer,
                                     reason: HLSSegmentProducer.PumpExitReason) {
         // #65 (VOD only): a broken backpressure wedge means AVPlayer is stuck behind a parked producer.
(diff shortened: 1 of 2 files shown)


superuser404notfound:
Merged. Verified onto current main (the branch was based on 6.15.3): 1689 swift-testing plus 483 XCTest green, tvOS build clean. One conflict on the way in, since #341 landed `liveRecoveryBudgetDecision` at the same spot; both pure decisions now sit next to each other.

The analysis holds all the way down. A `.url` source reaches both exhaustion sites with its provider un-halted precisely because the transport exists: `shouldHaltLiveProduction` returns false for the recoverable exit reasons, so the barren-cycle cap and the attempt cap are the only places left to latch it, and #199 wired only the transport it introduced. Since #342 a frozen URL live session does get a consumer-side rescue (the stall ladder publishes `liveSourceReset` on the final rung), but that arm never halts production, so the dead provider kept advertising blocking reloads it could not answer. This closes the producer-side half, which is the half that owns the latch.

Two things pulled in on top, in 23a50655.

`transport != .none` classifies any transport added later by accident, and `.none` is the standing proof that "everything except one case" was already the wrong shape once: a future transport that delegates to host retune before reopening (exactly what `.none` does) would inherit the escalation and signal the host twice. It is an exhaustive switch now, so the compiler asks instead.

Both exhaustion sites also carried their own copy of the two effects, and the halt is the half that is easy to lose. One method owns the decision and both effects now, and the two new tests go through that method on a real engine and provider, so a call site that regresses fails them where the predicate-only pins could not. Checked by mutation: dropping the halt fails the URL pin, removing the guard fails the `.none` pin.

Not released yet, it goes out with #340, #341 and #342.

Issue #232, opened 2026-07-27
Field-order routing sends progressive-in-interlaced-carriage (PsF) H.264 to software for no visual benefit — #220 reprices the stated cost


rrgomes:
`VideoRoutingPolicy` routes H.264 to software when the container or avcC declares a field order, so `DeinterlaceFilter` can run — tvOS AVPlayer will not deinterlace, and 1080i broadcast content otherwise combs. The rule is right, and its comment prices the false positive honestly: *"A false positive only costs an unnecessary SW decode (deint=interlaced passes progressive frames through untouched), never a wrong deinterlace."*

There is a systematic false positive, and #220 has changed what it costs.

## PsF

Blu-ray has no 1080p25, so 25 fps European content is mastered as 1080i25: progressive pictures in interlaced carriage. The container and SPS declare it interlaced and they are not lying about the encoding — but the pictures are progressive, and nothing needs deinterlacing.

On a 46-minute 1080i25 H.264 episode from a European Blu-ray remux (DTS-HD MA 5.1, one PGS track):

```
ffprobe stream:  field_order=tt          container and SPS declare TFF interlaced
ffprobe frames:  interlaced_frame=0      every sampled frame, I/P and B
```

MediaInfo agrees with the container — `Scan type: MBAFF / InterleavedFields / TFF`, SEI `pic_struct=3` — while every decoded frame comes out with equal field POCs and unflagged. Both tools are right at different layers.

The engine's own behaviour confirms it end to end. `SoftwareVideoDecoder` engages the filter only on flagged frames, so the graph never configures: logs from these sessions carry only the 64×64 warm-up probe and never a frame-size `[Deinterlace] engaged` line. The picture is correct, and the graph is right not to build. The content is decoded in software, and then nothing deinterlaces it.

## What changed

Until now the cost of this false positive was one unnecessary software decode. #220 added a second cost: the software path with a subtitle track selected is where that issue lives, so the class of content that trips this rule for no benefit is also the class exposed to it. Independently, software-decoding 1080i25 H.264 alongside DTS-HD MA forgoes VideoToolbox and spends power and thermal headroom for a filter that never runs.

## Proposal

[...]

Issue #337, opened 2026-08-08
Software path: cold-start clock arming deadlocks when the selected audio stream first appears after the renderer fills (state=playing, currentTime=0 forever; one seek unwedges)


edde746:
### What happened?

A software-path VOD session can come up with `state == .playing`, frames decoded, the layer `isReadyForDisplay`, `hasFirstFrameReadyForDisplay == true` — and `currentTime` pinned at 0 forever. No error, no timeout, nothing for a host to react to. One seek unwedges it and playback is normal from there.

Trigger: `selectAudioTrack(index:)` shortly after `play()`, selecting a track whose first packet sits **later in the mux than the point where the video renderer fills**. The audio-switch reload rebuilds the host with `resumeAt = 0.00`, and the rebuilt session deadlocks on its own cold-start clock arming.

Field case: Apple TV 4K (A15, tvOS 26), AV1+Opus MKV via Plex, host applies the viewer's preferred audio language ~20 ms after `play()` (engine default was track 1/jpn, preference was track 2/eng). Screen shows the first frame, position never moves; the viewer's first seek starts playback. Verbatim from the device log:

```
[AetherEngine] reload begin: audioStream=2 resumeAt=0.00s embeddedSub=-1 sidecar=nil
[AetherEngine] reload: loadSoftware done (414ms)
[AetherEngine] reload: state=.playing total=420ms
[SWHost] first video frame enqueued: pixfmt=0x34323076 size=1916x1072 pts=0.000s
[Renderer] enqueue #1: status=rendering ready=true error=nil
[SWHost] layer.isReadyForDisplay=true after 5 frames
   ... 19 s of state=playing with the host's progress reports stuck at time=0 ...
[AetherEngine] seek#1 programmatic began target=10.00
[AudioOutput] seekClock to=10.000 rate=1.0     <- first time the clock ever moves
[AetherEngine] seek#1 programmatic landed rendered=10.00 target=10.00
```

### The deadlock

Four facts, each individually deliberate, that close a cycle (all refs 6.15.2, `Native/SoftwarePlaybackHost.swift`):

1. `play()` deliberately does not start the clock cold: "demux loop arms the clock on first decoded audio sample" (`play()`, ~716).
2. Arming happens in the demux loop's **selected-stream** audio branch, on the first decoded buffer (~1855).
[...]

PR #346, opened 2026-08-09
Sequential-origin sources: one unranged connection, native routing, and an append playlist with real durations

tschuegy:
## Problem

IPTV timeshift/catch-up archives are served by origins that **fabricate HTTP range answers**: any `Range: bytes=X-Y` gets a plausible-looking `206` + `Content-Range: bytes X-Y/total`, but the body is positioned on a coarse internal chunk boundary rather than byte X. Live probes against a real panel showed a request for offset 524288 returning the bytes of offset 0, suffix ranges returning the *first* N bytes, and range ends ignored outright. Only byte 0 is addressable, deterministically.

Through the existing reader this produced three device-visible failures on catch-up playback:

1. The 32 MB range rotations spliced ~1.9 s of misplaced content into every reconnect (mpegts `Packet corrupt`, negative audio packet durations, audio desync roughly once a minute).
2. FFmpeg's tail-read duration estimate ran against fabricated offsets and probed a 135-min window as 9.5 h.
3. Even with a clean sequential read, the static VOD plan's uniform `EXTINF:4.000` was a lie for archives whose GOP cadence does not divide the cut target: a 1.92 s-GOP channel cut eleven 3.84 s segments and one 5.76 s segment per 48 s cycle, all advertised as 4.000, and AVPlayer on tvOS 26 visibly jumped ~1.8 s at every resync while every queue metric read clean. (A sibling channel whose keyframes happen to sit on the 4 s grid played the same pipeline flawlessly - that A/B is what finally pinned it.)

## Change

Four commits, each standing on the previous:

- **`fix(io)`** - `LoadOptions.sequentialOrigin` declares the shape (headers cannot expose it, so it is a caller declaration, not a probe): the reader routes onto its existing forward-only streaming mode - one long-lived unranged GET, no suffix/tail probes, no size probe, no detours, no byte-offset reconnects, pb non-seekable. A connection that dies short of its advisory Content-Length reports **EIO instead of EOF** (EOF reads as end-of-media, which hosts deliberately never retry), and the #169 VOD readError revive is not admitted (its fresh demuxer could only reopen from byte 0 and then fails its anchor seek, burning a slot on connection-capped panels). The paired `declaredDurationSeconds` carries the caller-known window length with the highest precedence in the AE#105 duration chain. `aetherctl play` grows `--sequential-origin` / `--declared-duration`.
- **`fix(routing)`** - a sequential origin keeps the **native path**: every reason the forward-only rule forces software (cue prewarm, segment seeks, size probes) is already answered for the explicit declaration, and the producer reads the archive linearly from byte 0 exactly like a live session does. Declared-interlaced archives still route software via the field-order policy (the #232 refute probe stays seekable-only: its sample cannot be rewound on a sequential reader).
- **`fix(hls)`** - sequential sessions render their media playlist the way the live path always has: **only finalized segments, each with the duration actually muxed**, as an append-only EVENT playlist that completes with ENDLIST at true source EOF. The server holds the first playlist until the startup segments exist (mirrors the live gate). Two host-facing seams from the growing playlist are closed at the engine: a caller-declared duration outranks the item's, and a sequential session with no resume anchor queues a post-readiness seek to 0 through the #127 replay (AVPlayer defaults an EVENT playlist to edge-minus-holdback and re-anchors there after the load-time seek).
- **`fix(hls)`** - the finalize reports pair capture and duration **order-independently** (the muxer rotation can fire off an audio packet crossing the boundary before the first video packet of the new segment has recorded its start), plan indices a long GOP skipped report as zero-duration holes the renderer omits, and a small reorder buffer keeps the provider's strictly-contiguous append API satisfied.

## Verification

- `SequentialOriginReaderTests` (7 tests) against an extended `ThrottledOriginServer` (now logs Range-header *presence* - a range-less GET was indistinguishable from `bytes=0-` - and gains a `serveThenDrop` directive for the short-body EIO case): exactly one unranged request past the 32 MB rotation point, EIO on short body, EOF on clean end, profile plumbing, duration precedence.
- Full suite green (1638 tests).
- Device (tvOS 26.6, Apple TV 4K, real panel): a 720p50 archive and a 1080i-sourced 1.92 s-GOP archive both play clean end-to-end; the served playlist reads `EXTINF:5.760/3.840` matching the muxed media exactly; start lands at 0; the scrubber shows the declared window length.

The sequential declaration deliberately leaves scrubber seeking unavailable (byte seeking never worked against a lying origin - it was silently wrong); the follow-up story for these sources is app-side re-request with a shifted Xtream start timestamp.

### DIFF
--- a/Sources/AetherEngine/AetherEngine+Loading.swift
+++ b/Sources/AetherEngine/AetherEngine+Loading.swift
@@ -126,7 +126,15 @@ extension AetherEngine {
         }
         duration
             .sink { [weak self] value in
-                if value > 0 { self?.duration = value }
+                guard let self else { return }
+                // A caller-declared duration outranks the host's: a sequential session's append
+                // playlist grows while it plays, so the item duration is the produced span, not
+                // the window length the host UI should scale its scrubber to.
+                if let declared = self.loadedOptions.declaredDurationSeconds, declared > 0 {
+                    self.duration = declared
+                } else if value > 0 {
+                    self.duration = value
+                }
             }
             .store(in: &cancellables)
         isReady
@@ -1125,6 +1133,14 @@ extension AetherEngine {
         // forwardBufferDuration default (4 s): deep buffer lets AVPlayer race to the live edge and hit the transcode warm-up gap head-on (-12888); 4 s PACES consumption. Verified: 8 s worsened startup pause (8-10 s vs ~1 s).
         // Live REJOIN: skip initial seek so AVPlayer picks edge-minus-holdback instead; seek-to-0 against the re-served backlog wedged the reloaded item in waitingToPlay (device repro: tvOS 26, Jellyfin stream.ts). See LiveReloadPolicy.
         lastNativeVideoStartPosition = startPosition ?? 0
+        // Sequential append playlist: AVPlayer treats the growing playlist as an EVENT and
+        // defaults to edge-minus-holdback (~6 s in on a fresh session, more once the producer
+        // has raced ahead). The load-time seek to 0 fires before readyToPlay and the item
+        // re-anchors to the edge default afterwards, so queue a post-readiness seek through
+        // the #127 replay instead - every segment stays retained, so 0 is always reachable.
+        if !isLive, loadedOptions.sequentialOrigin, startPosition == nil {
+            pendingPreReadySeekSeconds = 0.0
+        }
         // AE#158: consume-and-reset so only the load() that armed the handover swaps in place; audio-switch
         // and recovery reloads keep their own contracts.
         let inPlaceHandover = pendingInPlaceItemHandover
--- a/Sources/AetherEngine/AetherEngine+Loading.swift
+++ b/Sources/AetherEngine/AetherEngine+Loading.swift
@@ -502,6 +502,8 @@ extension AetherEngine {
             // Caller-bounded probe budget (#68) for the fallback open / live reopen; the happy path reuses preopenedDemuxer.
             probesize: loadedOptions.probesize,
             maxAnalyzeDuration: loadedOptions.maxAnalyzeDuration,
+            sequentialOrigin: loadedOptions.sequentialOrigin,
+            declaredDurationSeconds: loadedOptions.declaredDurationSeconds,
             forwardBufferSegments: loadedOptions.forwardBufferSegments
         )
         // #240: the pump claims the source link through this gate while it is fetching, so the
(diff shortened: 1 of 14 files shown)


superuser404notfound:
Merged, and verified on current main first. The only conflict was in aetherctl, where `--switch-audio` (fb2ea4c4, from the #337 work) had landed on the same signatures; both flag sets are kept and the engine changes went in untouched. 1706 swift-testing plus 483 XCTest green here.

The analysis holds, including the part that is easy to get wrong: this cannot be probed. An origin that answers every range with a well-formed 206 is indistinguishable from a correct one at the header level, so a declaration is the only honest shape, and pairing it with `declaredDurationSeconds` is right because the tail estimate goes out with the rest of the ranged reads. The A/B against the sibling channel whose keyframes sit on the 4 s grid is what makes the third failure legible; a uniform EXTINF over a cadence that does not divide the cut target is exactly the kind of lie every queue metric reads clean.

Some follow-up hardening landed on top in f09a0415.

The byte-0 rule has three call sites, and the revive was one of them. `performRestart` seeks the demuxer and does not treat a failed seek as failure, so on a non-seekable pb the new producer keeps reading wherever the stream stands (or from byte 0 after a fresh reopen) and labels those bytes as segment `idx`. That is the same fabricated-position content the declaration exists to keep out, except silent instead of audible, and two paths reach it on a sequential session: the provider's restartHandler when a requested segment falls outside the producer's coverage, and the seek-deadline re-anchor, which is authoritative and so wins the coalescer. The third site is the #93 resume anchor: a first producer at idx > 0 emits reports that are not contiguous from 0, the append playlist drops every one of them by contract, and the startup gate times out on an empty asset. One predicate owns the rule now and all three sites ask it. Worth knowing on device: a restart request now ends the session through `onVODSourceFailed` rather than producing mislabeled segments, which is the loud version of a failure that used to be quiet.

The EOF tail EXTINF could never reach its real span. `lastMuxedItemAxisSeconds` was recorded at the ledger site, which only fires when a segment opens, so it was always equal to that segment's own start and the `> start` branch was dead: every archive's final segment was advertised at the full cut target regardless of how little media it held. It is recorded per packet now.

The zero-duration skip in the playlist renderer was reaching the live path too. Dropping a URI shifts every later segment's implicit media sequence number, which is the number a blocking reload resolves against, so it is scoped to the append playlist and there is a pin on the live renderer still emitting a zero-duration entry. Checked by mutation: with the old form that pin fails and seg1 disappears from a live playlist.

Also moved AVERROR(EIO) out of its private constant in AVIOReader into `FFmpegErr.eio`, next to `eof`, since being distinguished from `eof` is the entire point of it.

Thanks for this one. The declaration plus append-playlist shape is the piece that makes these archives playable at all, and the device A/B is what carried it. Not released yet, it will go out with #340 through #343.

Issue #344, opened 2026-08-09
visionOS 1.0 minimum conflicts with 1.1-only AVFoundation APIs


YangHanqing:
### What happened?

AetherEngine 6.15.3 fails availability checking when built for visionOS.

Package.swift declares `.visionOS(.v1)`, but the engine references AVFoundation APIs introduced in visionOS 1.1 without a visionOS availability check:

- `AVSampleBufferVideoRenderer.videoPerformanceMetrics`
- `AVSampleBufferDisplayLayerReadyForDisplayDidChange`
- `AVSampleBufferDisplayLayer.isReadyForDisplay`

The surrounding checks name tvOS, iOS, and macOS, then fall through to `*`. For visionOS, that allows 1.0, so these references are rejected by the compiler.

My app targets visionOS 26, but the package's own declared minimum still makes this a compile-time error.

### Steps to reproduce

1. Add AetherEngine 6.15.3 to a visionOS target.
2. Build the target with the visionOS 26 SDK.
3. Observe availability errors for the 1.1-only APIs listed above.

No media is involved; this fails at compile time.

### AetherEngine version or commit SHA

6.15.3 / 643f0322ac02cd7228f6df93f6fb55907554b5af

### Host app

Custom / my own integration

### Platform

iOS

### OS version

visionOS 26.0 SDK (xrsimulator)

### Device / chip

Apple Vision Pro simulator, arm64

### Playback path

Not sure / not a playback bug

### Source media (for playback bugs)

N/A — compile-time availability issue

### Error codes / log lines

```text
'videoPerformanceMetrics' is only available in visionOS 1.1 or newer
'AVSampleBufferDisplayLayerReadyForDisplayDidChange' is only available in visionOS 1.1 or newer
'isReadyForDisplay' is only available in visionOS 1.1 or newer
```

### Anything else

Suggested fix: raise the package minimum to `.visionOS(.v1_1)`. If visionOS 1.0 support is intentional, each use needs an explicit `#available(visionOS 1.1, *)` guard instead.

The form does not offer visionOS as a platform, so I selected iOS above.

superuser404notfound:
[...]

PR #347, opened 2026-08-09
SW path: fold PTS discontinuities on forward-only VOD, decouple audio decode-ahead from the video gate

tschuegy:
## Problem

Two SW-path weaknesses surfaced while chasing an IPTV catch-up bug on device (tvOS 26.6), both independent of the source that exposed them:

1. **The live PTS-discontinuity fold is gated `isLive`, and non-live sessions run unprotected.** A forward-only VOD source (chunked timeshift archive) restarted its timestamps at PTS ~0 every ~85 s; FFmpeg's 33-bit wrap correction read the -2718 s jump as **+92726 s**, the renderer waited 25 hours for the frame's display time, and the video queue died with `FigVideoQueueRemote -12080` - picture and sound frozen ~90 s into every session.
2. **The combined demux loop paces everything on the video renderer's ~10-frame queue**, so interleaved audio can never build more than ~0.3 s of lead over the synchronizer clock (the master clock). Any decode/deinterlace jitter beyond that starves the audio renderer and the clock leaps; a timeline gap lets the free-running clock run away while the picture freezes. The DVR feeder arm has `AudioLookaheadPolicy` for exactly this; the combined loop had nothing.

## Change

- **`fix(sw)` fold** - the discontinuity fold gate widens to the new pure decision `shouldFoldTimeline(isLive:sourceSeekable:)`: live folds as before, and a forward-only non-live source folds too (its timeline is served as-is and a non-seekable pb offers no seek-based recovery). Seekable VOD keeps its trusted container timeline untouched. Seams log per occurrence with a soft cap of 20 lines.
- **`fix(sw)` audio decoupling** - video packets park in a bounded FIFO (drained at the renderer's pace) while audio keeps decoding up to `AudioLookaheadPolicy.targetLeadSeconds` ahead of the clock, and a genuine audio underrun **pauses the clock for a rebuffer** via the same policy the DVR feeder uses, with an `everHadLead` latch so exactly-realtime origins do not eat a spurious pause at session start. At a fold seam the parked pre-seam tail is decoded before the flush drops its reference chain. Live-without-ring sessions keep the historical lockstep pacing.
- **`diag(sw)`** - the SW path gains a 1 Hz `[SWDiag]` line (clock + delta, decoded-audio lead, parked depth, rebuffer state, the display layer's own drop counter, accumulated render delay with per-second delta, queue-target status, surface state, `isReadyForDisplay`). The native path has LagDiag; SW sessions only had the 30 s memprobe, which is too coarse to see the clock leaps and layer-drop bursts a stuttering session is made of. It also names a headless-harness artifact: a never-bound layer (`surf=DETACHED`) drops ~8 frames/s that a bound one does not.

## Verification

- `SWTimelineFoldTests` pins the fold decision; full suite green (1632 tests).
- Device: with the fold active, the chunked archive's seams fold cleanly instead of freezing into -12080; `[SWDiag]` on a healthy session reads `dclk=1.00`, `aLead` 3-5 s sawtooth, `layerDrop` flat, `delay=0.00`.
- On an M-series Mac via `aetherctl play`: clock exact across 5+ minutes, no rebuffer churn on fast origins, rebuffer pause/resume behaves on origins that deliver below realtime.

### DIFF
--- a/Sources/AetherEngine/Native/SoftwarePlaybackHost.swift
+++ b/Sources/AetherEngine/Native/SoftwarePlaybackHost.swift
@@ -1146,6 +1146,7 @@ final class SoftwarePlaybackHost {
             return
         }
 
+        let diag = demuxDiag
         demuxQueue.async {
             Self.runDemuxLoop(
                 demuxer: dem,
@@ -1158,6 +1159,7 @@ final class SoftwarePlaybackHost {
                 condition: condition,
                 initialClockTime: initialClock,
                 currentRate: currentRate,
+                diag: diag,
                 ring: ring,
                 videoTimeBaseSeconds: vTbSec,
                 audioTimeBaseSeconds: aTbSec,
@@ -1631,6 +1633,7 @@ final class SoftwarePlaybackHost {
         condition: NSCondition,
         initialClockTime: CMTime,
         currentRate: @Sendable () -> Float,
+        diag: SWPlaybackDiagState? = nil,
         ring: PacketRingBuffer?,
         videoTimeBaseSeconds: Double,
         audioTimeBaseSeconds: Double,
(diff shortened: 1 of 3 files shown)


superuser404notfound:
Merged, thanks. Both halves land where the defect is: the fold decision is a property of the timeline the source serves, not of the flag the session was started with, and the audio pump belongs in the engine rather than in whatever the host can compensate for. `[SWDiag]` is the part I expect to keep paying off, since software sessions had nothing between the 30 s memprobe and a device capture.

Merged in 1fec4668 (main had moved since the branch opened: `presentationSize` from #353 and `armFromVideoPacket` from #337 sat on the same two spots, both conflicts additive). Hardening followed in 8013eefc:

- **The read gate paces on the lead again.** `AudioLookaheadPolicy.targetLeadSeconds` was named in the comment but never read: the only thing that stopped the loop was the 256-packet cap, so the effective lead was "however many seconds of video fit in 256 packets" and moved with frame rate (~10 s at 25 fps, ~5 s at 50 fps, which is roughly why the measured sawtooth landed near 4 s on 1080i50). `shouldHoldDemuxRead()` asks the policy; the cap stays as the memory backstop it was meant to be.
- **One method owns waiting on the renderer.** A rebuffer hold stops the synchronizer, and a stopped synchronizer never drains the renderer, so all three waits (cap gate, fold seam, end of media) could sit under a hold that only this same thread could lift. The DVR arm may wait because its pump runs on a second thread; the combined loop may not. `waitForRenderer()` lifts the hold before it waits.
- **#337 reaches the parked path.** The unarmed-clock exit lives in the lockstep gate the park bypasses, and the park's own fallback needs `audioPacketsSeen >= 50`, which a stream producing nothing never reaches. Same exit, anchored on the FIFO head.
- **Parked packets are seek-generation checked before decode**, the way the lockstep gate re-checks after its wait (a pre-seek packet clears the decoder's skip threshold).
- **`everHadLead` resets with the seek.** The lead is zero again afterwards, so keeping the latch set pauses the clock for a rebuffer on the first post-seek check, which is what the latch exists to prevent.
- The `[SWDiag]` availability list needed `visionOS 1.1` for `isReadyForDisplay`. That is #344 again: `*` resolves to the declared visionOS floor (1.0), not "anything newer", and CI now builds that platform.

`SWDemuxReadGateTests` pins the gate decision; full suite green at 1712 tests, visionOS Simulator build green. Not tagged yet, this rides along with the rest of the series.

superuser404notfound:
Shipped: the whole 340 to 347 series is in [6.19.0](https://github.com/superuser404notfound/AetherEngine/releases/tag/6.19.0), tagged at fecc765a. Both downstream apps are already pinned to it, so a retest on device has a released engine to sit on now.

Most interesting to hear about, in rough order: whether the chunked archive still seams cleanly over a long session (the fold is unbounded in offset accrual, so a multi-hour catch-up run is the real test), what `[SWDiag]` reads on a session that stutters rather than a healthy one, and whether the live arms hold up on a channel that reconnects a lot. `aLead` should now sit at the 4 s target rather than wherever the FIFO cap put it.

PR #349, opened 2026-08-09
A play gate that saw the switch start waits for its end (Sodalite#49)

superuser404notfound:
Follow-up to #345, on the configuration Sodalite#49 was originally filed on: a rate-only write, which takes no pre-flight at all.

## Measurement

Apple TV 4K (3rd gen), tvOS 26.5, SDR 25p on a 60 Hz system, two runs:

| | run 1 | run 2 |
|---|---|---|
| item ready | +443 ms | +286 ms |
| gate released `play()` (cap) | +2130 ms | +2068 ms |
| **panel finished switching** | **+3555 ms** | **+3550 ms** |

Two things this shows. A **rate** switch on this panel takes ~3.55 s, longer than the ~2.85 s of a dynamic-range switch, so the earlier figure was the range handshake rather than a panel constant. And the gate released 1.4 s early, because `stage2CapMs` is sized for a gate that runs *before* the load, where breaking out early is what releases `loadNative` (see the note on that constant, and AE#348).

The part no log shows: **the panel is dark for the whole switch.** The reporter confirmed it on the device. So the early release bought no picture at all, the first visible frame lands at +3555 ms either way, and the 1.4 s it "saved" were 1.4 s of the film played into a black screen.

## Change

A gate that runs after the item is ready blocks nothing else, so when it has recorded the switch starting it waits for the end, up to a 6 s ceiling (measured switches: 2779 to 2898 ms range, ~3550 ms rate, none came close).

Unchanged, deliberately:

- **the pre-flight**, where the early break-out is load-bearing until AE#348 makes the load concurrent
- **live**, where a zap must not sit behind a panel handshake
- **a switch with no recorded start**, the unobservable-DV panel: nothing will report an end there, so a longer ceiling would be pure startup cost

Because the wait ends on the observed end rather than on a duration, a panel that switches quickly pays only what it takes.

## On "a rate-only switch does not need to be waited out"

Raised on the Sodalite#49 thread. Correct as far as failure modes go: only a range change can fail AVPlayer outright. But the premise behind it, that waiting costs startup time, does not hold on a panel that blanks: there the wait is free and skipping it costs content instead. On a panel that switches without blanking the trade is the other way round, and this change is sized for that too, since it waits for the end rather than for a number.

## Verification

`swift test`: 1655 tests, green. tvOS build green. Device verification of this change is pending; the measurements above are from the build before it.

### DIFF
--- a/Sources/AetherEngine/AetherEngine+Loading.swift
+++ b/Sources/AetherEngine/AetherEngine+Loading.swift
@@ -1674,12 +1674,14 @@ extension AetherEngine {
                 // #274: this reload preserved the criteria (resetDisplayCriteria: false) and re-applies none,
                 // so only a sole-writer host's re-write on the swapped item can still switch anything; the
                 // published (panel-clamped) format decides whether that can be a dynamic-range switch.
-                await displayCriteria.waitForSwitch(startGrace: Self.playGateGrace(
-                    criteriaUnchanged: false,
-                    engineIsCriteriaWriter: !loadedOptions.suppressDisplayCriteria,
-                    formatKnown: true,
-                    effectiveFormat: videoFormat
-                ))
+                await displayCriteria.waitForSwitch(
+                    startGrace: Self.playGateGrace(
+                        criteriaUnchanged: false,
+                        engineIsCriteriaWriter: !loadedOptions.suppressDisplayCriteria,
+                        formatKnown: true,
+                        effectiveFormat: videoFormat
+                    ),
+                    settleCap: loadedOptions.isLive ? .standard : .awaitObservedEnd)
                 try checkLoadCurrent(gen)
                 nativeHost?.play()
             }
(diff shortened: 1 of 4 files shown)


superuser404notfound:
Device-verified on the shipped 6.16.0 build (Apple TV 4K 3rd gen, tvOS 26.5, SDR 25p rate-only, three runs):

| run | item ready | switch ended | gate released `play()` |
|---|---|---|---|
| 1 | +438 ms | +2858 ms | +2887 ms |
| 2 | +306 ms | +2857 ms | +2861 ms |
| 3 | +418 ms | +2880 ms | +2915 ms |

All three release 4 to 35 ms after the panel finished, against 1425 and 1482 ms *before* it on the build this change was measured on. The item is ready around +400 ms in every run, so the gate is carrying the whole wait alone, which is exactly the case the standard cap was mis-sized for.

One thing worth recording: the rate switch measured 2835 to 2880 ms here against ~3530 ms in the earlier round, same panel and same target mode. The spread is wider than the dynamic-range switch ever showed, which is an argument for the shape of this change rather than against it: the gate now ends on the observed end, so a switch that takes 2.8 s costs 2.8 s and one that takes 3.5 s costs 3.5 s. Any fixed number would have been wrong for one of those two rounds.

Issue #370, opened 2026-08-13
Sequential EVENT startup: the playlist is held for 2 published durations (= 3 segment opens) and the spacing scan spends the origin's non-replayable prefix


tschuegy:
## Symptom
On a slow or stalling origin, a sequential-origin session (AetherEngine 6.25.1, tvOS 26.6, Xtream timeshift `.ts`) never starts: AVPlayer's first `/media.m3u8` GET is held for the full 30 s startup wait and the asset load times out (`NSURLError -1008` / CoreMedia `-12884`) even though two segments were already produced and ~12 s of media sat on disk:

```
[HLSSegmentProducer] seg-0.m4s captured (2336504 B)
[HLSSegmentProducer] #65 ledger seg-1 base=0 itemAxis=6.000s sourceStart=32076.800s planSource=32074.800s drift=2.000s shift=32070.800s
[HLSSegmentProducer] seg-1.m4s captured (2308798 B)
... 30 s later ...
[NativeAVPlayerHost] #8 asset.load(duration) failed: NSURLErrorDomain/-1008 ... underlying=CoreMediaErrorDomain/-12884
[HLSLocalServer] media.m3u8 tail: ... #EXTINF:6.000, / seg0.mp4
[HLSLocalServer] -> 200 /media.m3u8 bytes=176
```

## Cause — two compounding startup costs
1. **The gate demands 2 published durations, which takes 3 ledger opens.** `waitForSequentialStartupSegments` reuses `LiveEdgePolicy.minStartupSegments = 2`, a live sliding-window constant (its doc comment guards against `-12888` on a window whose holdback can't fit). On the append-only EVENT playlist a segment's EXTINF only becomes final when the *next* segment's ledger opens (real duration = nextStart − start), so N published durations require N+1 segment opens: the gate effectively waits for ~3 segments ≈ 12-18 s of media through a possibly-throttled origin. A one-segment EVENT playlist is legal HLS, and this playlist already carries the refresh counter tag that defeats AVPlayer's unchanged-playlist patience — the live rationale does not apply.
[...]

Issue #369, opened 2026-08-13
Wrap-escape containment: unbounded inferred sample duration, discarded discontinuity-scale fold runs, and a backpressure park that waits on an unadvertisable index


tschuegy:
## Context
Companion to the sequential chunk-seam rebase issue (#368): even with the rebase in place, a timestamp leap that escapes it (corrupt stream, a jump inside one chunk, future regressions) currently turns the session into a 4-minute zombie instead of failing fast. Three independent containment gaps, all observed in one device trace (AetherEngine 6.25.1, tvOS 26.6, sequential-origin Xtream timeshift):

## 1. An inferred sample duration longer than a discontinuity is accepted
`resolveVideoSampleDuration` returns `nextDts − dts` unbounded. Across a 2^33 wrap the look-behind finalize computes 8 226 410 192 ticks (~91 404 s @ 90 kHz) and hands it to movenc:

```
[mp4 @ 0x15eb6e700] [error] Application provided duration: 8226410192 in stream 0 is invalid
```

`finalizeAndWriteVideo` then discards the write return code, so the packet is silently lost and nothing in the log names the failure at ERROR level from our side.

## 2. Discontinuity-scale fold runs are discarded, disarming the #358 recovery arms
`SegmentCache.noteFolded` drops runs wider than `maxFoldRunLength = 64` ("a jump this wide is a restart or a seek, not a fold"). The field case was precisely such a run — 312 indices:

```
[HLSSegmentProducer] #358 plan indices 2...313 folded into seg-314 (no IRAP reached their boundary)
```

With the counters at 0, both #358 recovery arms (`foldedTargetDecision` reanchor and the engine's unrecoverable-gap handler) stay disarmed, so the designed fold-once-reanchor / fold-twice-fail path never runs and the session limps on until the #65 wedge breaker trips minutes later.

## 3. The backpressure park can wait on an index the playlist can never advertise
The advance-path park waits for the consumer to fetch `head − bufferAheadSegments`, but on a sequential EVENT playlist the advertised frontier is fed by this same pump's finalize reports — zero-duration fold holes get no URI. After a fold-to-tail the park waits for a fetch of an index that will never appear:

```
[HLSSegmentProducer] #65 backpressure PARK (advance) head=374 target=364 cacheTarget=54 highStored=61 cached=62 parked=12s (no playback progress)
```

[...]

PR #371, opened 2026-08-13
fix(video): rebase the timeline across an archive chunk seam on a sequential origin (#368)

### LINKED ISSUE #368: Sequential-origin VOD: a 33-bit PTS wrap at an archive chunk seam is never rebased — cutter folds to the plan tail, playlist freezes, session dies on a refused re-anchor
## Environment
- AetherEngine 6.25.1, native loopback-HLS route, `LoadOptions.sequentialOrigin = true`, `declaredDurationSeconds` set (IPTV catch-up/timeshift `.ts`, forward-only, readable from byte 0 only)
- Apple TV 4K, tvOS 26.6; source: Xtream `/timeshift/.../YYYY-MM-DD:HH-MM/chan.ts` (H.264 720p50 + AC-3, field-coded, open-GOP)

## Symptom
Playback runs cleanly for ~4 minutes, then the served EVENT playlist stops growing, AVPlayer drains its forward buffer, stalls, and the session dies terminally. A client-side retry of the same archive dies again ~2 s after start.

## Root cause
IPTV timeshift archives are chunked recordings; each chunk restarts near PTS 0. libavformat's 33-bit wrap correction turns the backward step at a chunk seam into a forward leap of +2^33. Device fingerprint from the retry session:

```
[mp4 @ 0x15eb6e700] [error] Application provided duration: 8226410192 in stream 0 is invalid
```

`363524400 + 8226410192 = 8589934592 = 2^33` exactly — the seam packet's item-axis DTS leaps ~91 404 s forward.

`HLSSegmentProducer` has no discontinuity handling for non-live sessions: both the timeline rebase and the raw-PTS discontinuity detector are gated on `isLive` (`HLSSegmentProducer.swift:2310`, `:2877`), and the monotonic gate only repairs ≤0.5 s backward glitches. The leap therefore reaches `VODSegmentCutter.index(pts:isKeyframe:)` unmodified, whose monotonic while-loop walks the segment index to the tail clamp:

```
[HLSSegmentProducer] #358 plan indices 2...313 folded into seg-314 (no IRAP reached their boundary)
```

From there the session is structurally dead:
1. The backpressure park target is derived from the head (`head − bufferAheadSegments`), and release requires the consumer to fetch that index (`SegmentCache.awaitFetchHighWater`). After the fold, head sits at the plan tail, but the EVENT playlist skips zero-duration folded entries — the release index can never be advertised. Permanent park, frozen playlist:

```
[HLSSegmentProducer] #65 backpressure PARK (advance) head=374 target=364 cacheTarget=54 highStored=61 cached=62 parked=12s (no playback progress)
[NativeAVPlayerHost] #2 errorLog code=-12888 ... 'Playlist File unchanged for longer than 1.5 * target duration'
```

2. AVPlayer drains the listed window and stalls; the #65 fast-path breaker exits the pump and the host re-anchors on the AVPlayer position — which a sequential origin categorically refuses:

```
[HLSSegmentProducer] #65 backpressure WEDGE BROKEN (advance) head=374 target=364 cacheTarget=60 parked=49s (fast path: fetch target + rendered clock both frozen); exiting pump for host re-anchor on AVPlayer position
[HLSVideoEngine] #65 backpressure wedge: re-anchoring producer to 243.68s -> seg60 (AVPlayer position) (attempt 1/5)
[HLSVideoEngine] restart at idx=60 refused: a sequential origin can only be read from byte 0, so the reposition would mislabel content; surfacing source failure
```

## Precedent inside the engine
`SoftwarePlaybackHost` already folds exactly these seams, unconditionally, for `isLive || !sourceSeekable` — its comment cites an IPTV timeshift archive whose wrap became +92 726 s (`SoftwarePlaybackHost.swift:1731-1737`). The producer path lacks the equivalent, so the two hosts behave differently for identical load options.

## Proposed fix
Extend the existing live timeline rebase (video `:2310` and audio `:2379` gates) to sequential-origin VOD (`sequentialOriginPinsProducerToZero`), reusing the live thresholds. The cutter is fed post-shift item-axis DTS, so after a rebase it keeps walking the same plan boundaries sequentially — no cutter, ledger, or playlist change needed, and EXTINF stays real. No `EXT-X-DISCONTINUITY` is wanted: the archive seam is content-continuous and the output timeline stays continuous after the rebase.

PR follows.

tschuegy:
Fixes #368.

## What happens today

On a sequential origin (`LoadOptions.sequentialOrigin`, IPTV timeshift archives), a chunk seam's PTS reset — wrap-corrected by libavformat into a +2^33 leap — reaches `VODSegmentCutter` as item-axis time and walks its monotonic index to the plan tail in one call. From there the session is structurally dead: the append playlist freezes (folded indices publish duration 0 and get no URI), the backpressure park waits for a fetch of `head − 10` that the playlist can never advertise, AVPlayer drains and stalls, and the #65 wedge recovery's reposition is exactly what a sequential origin categorically refuses. Field trace: 4 minutes of clean playback, stall, and a retry session that died the same way ~2 s in (the arithmetic fingerprint: `363524400 + 8226410192 = 2^33` exactly, surfaced by movenc as `Application provided duration: 8226410192 in stream 0 is invalid`).

## The fix

Run the existing live timeline rebase (video and audio, unchanged thresholds and pairing machinery) for sequential-origin VOD as well: `rebasesTimelineOnDiscontinuity = isLive || foldsSequentialTimeline`, with the engine passing `sequentialOriginPinsProducerToZero`. The cutter, ledger, and append playlist need no change — they operate on post-shift output time, so after the rebase the cutter keeps walking the same plan boundaries sequentially and EXTINF stays real.

Deliberate choices, spelled out in comments:

- **No new LoadOption.** `sequentialOrigin` is already the client's declaration "this is a forward-only chunked recording"; timeline continuity across its chunk seams is a correctness property of that declaration. `SoftwarePlaybackHost.shouldFoldTimeline` already folds these seams unconditionally for `isLive || !sourceSeekable` (its comment cites the same failure at +92726 s), so gating the producer path behind a second flag would make the two hosts behave differently for identical options.
- **No `EXT-X-DISCONTINUITY` at the seam.** The archive is content-continuous and the output timeline stays continuous after the rebase; the discontinuity/force-cut flags stay live-only (they feed only the live segment bookkeeping).
- `onLiveTimelineRebase` fires for sequential seams too, on purpose: the A53 caption tap and subtitle mapping need the playlist axis moved the same way, and `handleLiveTimelineRebase`'s deferred-shift contract is exactly right for it.
- `isSourceReplay` stays untouched in the shared path: it requires a recent unplanned reconnect, which a sequential origin (single connection, no reconnect) can never have.

## Tests

- `TimelineRebaseMathTests` (new): the rebase math extracted as a pure static and pinned against the device trace — the 2^33 wrap continues exactly one fallback frame past the last output, a backward chunk restart (the SW-host trace) keeps output moving forward, a live-scale jump proves the extraction changed nothing for live, and the zero-fallback floor.
- `VODSegmentCutterTests`: the #368 contract from the cutter's side — rebased continuous dts advances one boundary per cut, while a raw wrap-scale leap saturates at the tail clamp in a single call.
- Full suite green on macOS: 1827 Swift Testing + 509 XCTest, 0 failures.

## Test plan

- Device: Apple TV 4K, tvOS 26.6, Xtream timeshift archives of SRF channels (H.264 720p50; AC-3 field-coded open-GOP on one, AAC on the other) — the two field sessions that die today. Verification runs through the Syravo client after the next release is pinned; both failure signatures (4-minute freeze, instant retry death) are reliably reproducible there today.

Merges cleanly alongside the #369 and #370 PRs; the three-way combination is integration-tested locally (full suite green on the merged tree).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,20 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Fixed
+
+- A sequential-origin VOD session now folds a timeline discontinuity the way live folds a
+  program boundary (#368). IPTV timeshift archives are chunked recordings whose every chunk
+  restarts near PTS 0; libavformat's 33-bit wrap correction turns that backward seam into a
+  +2^33 leap (device: dts delta 8226410192 ticks — 363524400 + 8226410192 = 2^33 exactly),
+  which reached the keyframe-gated cutter unmodified and walked its monotonic index to the
+  plan tail. After that the session was structurally dead: the playlist froze, the
+  backpressure park waited on a segment the playlist can never advertise, and the wedge
+  recovery's reposition is exactly what a sequential origin refuses. The existing live
+  rebase (both streams, same thresholds) now also runs for sequential-origin VOD; cutter,
+  ledger and append playlist need no change because they already operate on post-shift
+  output time. No `EXT-X-DISCONTINUITY` is added at the seam: the archive is
+  content-continuous and the output timeline stays continuous after the rebase.
 
 ## [6.25.1] - 2026-08-13
 
--- a/Sources/AetherEngine/Video/HLSSegmentProducer.swift
+++ b/Sources/AetherEngine/Video/HLSSegmentProducer.swift
@@ -170,6 +170,22 @@ final class HLSSegmentProducer: @unchecked Sendable {
     /// Live mode: cuts at keyframes past targetSegmentDurationSeconds; ignores segmentBoundaries.
     private let isLive: Bool
 
+    /// #368: a sequential-origin VOD session folds timeline discontinuities the way live does.
+    /// IPTV timeshift archives are chunked recordings; each chunk restarts near PTS 0, and
+    /// libavformat's 33-bit wrap correction turns the backward seam into +2^33 (device: dts delta
+    /// 8226410192 ticks, 363524400 + 8226410192 = 2^33 exactly). Without a rebase that leap reaches
+    /// the cutter as item-axis time and walks its index to the plan tail, after which the session is
+    /// structurally dead (frozen playlist, permanent park, refused re-anchor). The SW host already
+    /// folds these seams for forward-only VOD (SoftwarePlaybackHost.shouldFoldTimeline); this is the
+    /// producer-path equivalent. `isSourceReplay` stays in the shared path untouched: it requires a
+    /// recent unplanned reconnect, which a sequential origin (single connection, no reconnect)
+    /// can never have.
+    private let foldsSequentialTimeline: Bool
+
+    /// The timeline-rebase gate for both stream rebases: live program boundaries and sequential
+    /// chunk seams share one definition of "discontinuity" (the thresholds below).
+    private var rebasesTimelineOnDiscontinuity: Bool { isLive || foldsSequentialTimeline }
+
     private var liveCurrentSegmentIndex: Int
     private var liveSegmentStartPtsSeconds: Double = 0
     private var liveFirstSegmentOpened = false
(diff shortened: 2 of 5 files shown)


superuser404notfound:
Merged, thanks. The shape is right: the cutter, the ledger and the append playlist all work post-shift, so the rebase is the one place that needed to know about the seam, and reusing the live thresholds keeps one definition of "discontinuity" in the file. Verified here that `isSourceReplay` really is unreachable on this path, for a second reason beyond the single connection: the sequential reader reports EIO on a stall and never sets `lastUnplannedReconnectAt` at all.

One gap the rebase opens, fixed in a follow-up on main (20adae8e, merge a17a8b47). The published clock folds `item + shift - origin` against an origin latched once at session start, so moving the shift by the wrap moves the scrubber with it. Driving `applyNativeHostClockTick` across a seam assembled from the field traces (source start 32070.8 s, seam at item 244 s, wrap to 2^33 ticks) puts the playhead at 63378 s instead of 250 s on an archive whose declared duration is one hour, with `bufferedPosition` and `sourceTime` following it.

A sequential archive has no source axis to anchor a display origin on: every chunk restarts near PTS 0 and the wrap correction turns each seam into a fresh fiction, so no single source PTS maps to display-0 for the session. Its item axis does, it starts at 0 because the producer is pinned to byte 0, and it is exactly what `declaredDurationSeconds` measures. Such a session now publishes that axis; every other source keeps AE#270's latched origin and true source PTS.

Nothing structural depended on it, which is why the traces do not show it: the #65 wedge detector reads the rendered playlist position, not the engine clock. It is worth a look during the device retest though, since a host progress bar and any resume position ride on it.

Issue #358, opened 2026-08-10
VOD keyframe-aligned plan: the playlist advertises segment indices the producer never muxes, and the wedge recovery rebuilds the same hole


superuser404notfound:
Reported by @DrHurt through [AetherPlayer#2](https://github.com/superuser404notfound/AetherPlayer/issues/2). AetherPlayer 0.15.0 (engine 6.19.4), macOS 15.7.1, Mac mini 8,1 (x86_64, 6 cpus), local file. Playback stops about 18 s in and never comes back.

The source is a damaged mp4: `stream 41, missing mandatory atoms, broken header`, `stream 0, timescale not set`, `UDTA parsing failed retrying raw`, `Referenced QT chapter track not found`. HEVC 1920x1080, DV profile 8.1, 24.971 fps, 2536.7 s, 36 text tracks. The plan is the keyframe-aligned one: `766 IRAPs -> 603 segments [firstKeyframePts=-300000 (-3.000s)]`.

## The playlist advertises segments the producer never muxes

After a programmatic seek to 0 the producer restarts at idx 0 and the `#65 ledger` line reports these indices:

```
0  1  2  3  _  5  6  _  8  9  10  11  _  13  _  15
```

4, 7, 12 and 14 never appear. The ledger emits once per routed segment (`prevSeg != vodLedgerLastRoutedSeg`, `HLSSegmentProducer.swift:2858`), so a missing index was not muxed, it is a hole. `HLSSegmentProducer.swift:2866` already names the shape in a comment: "Plan indices a long GOP skipped report as zero-duration holes", but that reporting arm only runs when `onSequentialSegmentFinalized != nil`, which is the #346 sequential-origin path. On the ordinary VOD path the playlist keeps offering the index.

AVPlayer walks the playlist and dies on the first hole:

```
16:09:30.304  GET /seg0.mp4        ledger seg-0 .. seg-3 present
16:09:30.364  GET /seg1.mp4
16:09:30.389  GET /seg2.mp4
16:09:37.573  GET /seg3.mp4
16:09:44.890  GET /seg4.mp4        <- never produced
16:09:46.891  seg4: slow serve, sending early chunked header
16:09:48.832  playbackStalled  t+20.32s
```

Playback had resumed at 16:09:30.745, so it ran 18.1 s, which is exactly what the reporter sees.

## The plan and the muxed content disagree by up to 2.6 s

The same ledger lines carry the drift:

```
[...]

Issue #92, opened 2026-06-30
Native DV Profile 8.1 (HEVC) shows transient video glitches on the loopback-HLS path; not reproducible on replay


rrgomes:
**Symptom.** Playing native Dolby Vision **Profile 8.1** HEVC (single-layer, `dv_bl_signal_compatibility_id=1`, RPU present, no EL) on a DV display, brief blocky video distortion appears intermittently during normal playback:
- **Not reproducible on replay** -- seeking back to the same section does not reproduce it; glitch positions vary per play.
- **Absent in players that direct-decode without remuxing** (e.g. Infuse / VT+Metal) on the same file.
- **DV-P7 is clean** (libdovi RPU canonicalization + dvcC rewrite) and **non-DV 4K HEVC is clean** -- only native P8.1 glitches.

**Localization.** In `resolveCodecRoute`, native P8.1 on a DV panel is a pure passthrough (`convertP7ToProfile81=false`, `rewriteDoviConfigTo81=false`, `stripDolbyVisionMetadata=false`): the source RPU NALs (unspec62) + dvcC reach the mp4 muxer untouched, written as dvvC + SUPPLEMENTAL `dvh1.08.XX/db1p`. P7, by contrast, rewrites every RPU (`convertPacketToProfile81`, libdovi mode 2) and rewrites the dvcC.

**Negative result (useful).** I tried routing P8.1 RPUs through the same `convertPacketToProfile81` canonicalization P7 uses (set `convertP7ToProfile81 = effectiveDvMode` for `.profile81`). The conversions ran with **zero failures but the glitch persisted unchanged** -- so the raw RPU passthrough is *not* the cause. (It also added per-frame producer latency that made the seek wedge in the companion report more frequent, so I reverted it.) The differentiator is therefore something other than RPU content -- possibly the dvcC/dvvC signaling, or AVPlayer's handling of P8.1 DV over fragmented HLS at segment boundaries.

**Asset.** 4K (3840x1634) HEVC, DV Profile 8.1, E-AC3 5.1, MKV container, ~25 Mbps, ~54 min. The engine classifies it correctly:

```
[HLSVideoEngine] DV source: profile=8 compat=1 level=6 rpu=1 el=0 bl=1 color_primaries=9 color_trc=16 color_space=9
```

Happy to test patches or capture additional traces.

superuser404notfound:
Fixed on `main` (489e38c).

[...]

Issue #207, opened 2026-07-24
Forward-buffer window: 150-segment ceiling silently caps an "unbounded pre-buffer" host option (#102 follow-up)


fxndxs:
## Summary

`LoadOptions.forwardBufferSegments` (#102) made the loopback session's forward window configurable, but `HLSVideoEngine.clampedForwardWindow` caps it at **150 segments (~10 min at 4 s)**. A host that offers an explicit "buffer without limit" setting cannot honor it: any request above 150 clamps down, so a user who deliberately opted into pre-buffering the whole source for a flaky network (WAN/VPN streaming, commuting, trains/tunnels) still stalls once the connection drops for longer than ~10 minutes of content.

## Proposal

Raise the ceiling from 150 to **2700 segments (~3 h at 4 s)** — a pure sanity bound against accidental values that still covers a whole feature film. Hosts can then pass `Int.max` to mean "buffer the entire source". Floor (4) and nil default (10) stay unchanged, and 150 still passes through unchanged, so sessions that don't opt in are unaffected.

The cost profile stays the host's informed choice, as designed in #102: segments are disk-backed (mmap served, ~10 MB per 4K-HEVC segment — a fully cached 2 h film can occupy ~18 GB), and the producer front-loads the demux work. `SegmentCache` and `HLSSegmentProducer` take the window as plain arithmetic bounds (no fixed-size structures), so no other code change is needed; the backpressure/wedge machinery is untouched because the producer still paces against AVPlayer's fetch target plus the window.

If you'd rather keep 150 as the default ceiling, an alternative shape would be an explicit opt-in (e.g. `forwardBufferSegments: .unbounded`) — happy to rework the patch.

## Patch

[forward-window-ceiling.patch](https://github.com/user-attachments/files/30348577/forward-window-ceiling.patch)

Attached: `git am`-able patch against  5.20.4 (1abe7fa):

- `HLSVideoEngine.clampedForwardWindow`: ceiling 150 → 2700, doc comment explains the whole-film rationale and disk math.
- `LoadOptions.forwardBufferSegments` doc comment updated to match.
- `ForwardBufferWindowTests`: ceiling tests updated (900/2700 pass through, 2701/`Int.max` clamp to 2700); floor and nil-default tests unchanged.

### Motivating media or use case

[...]

PR #372, opened 2026-08-13
fix(video): contain a timestamp leap that escapes the timeline rebase (#369)

### LINKED ISSUE #369: Wrap-escape containment: unbounded inferred sample duration, discarded discontinuity-scale fold runs, and a backpressure park that waits on an unadvertisable index

tschuegy:
Fixes #369. Companion to #368's rebase: defense in depth for a timestamp leap that escapes it (corrupt stream, a jump inside one chunk, future regressions). Today such a leap turns the session into a multi-minute zombie; with these three containments it either keeps playing sanely or fails cleanly within seconds onto the designed `onVODSourceFailed` surface.

## The three containments

1. **`resolveVideoSampleDuration` gets a cap.** The look-behind delta across a 2^33 wrap IS the wrap (device: 8226410192 ticks ≈ 91404 s), which movenc rejects — and the write rc was dropped on the floor, so the packet vanished silently. The inferred delta now falls back (existing duration, then fallback) when it exceeds the discontinuity threshold: a sample longer than a discontinuity is definitionally invalid, and reusing `discontinuityThresholdSeconds` keeps one shared definition. The first rejected muxer write is now logged.

2. **`SegmentCache.noteFolded` counts wide runs.** The `maxFoldRunLength` guard dropped runs wider than 64 indices as "a restart or a seek" — but the field case was a discontinuity-scale cut leap (312 indices), and dropping it left every fold counter at 0, which is exactly what disarmed both #358 recovery arms for the folds most certain to trigger them. Wide runs now count (one Int per folded index, bounded by plan size); the constant survives as a log-classification threshold, and the fold site emits a distinct discontinuity-scale line. On a sequential origin the armed escalation reaches `requestRestart` → the sequential refusal → `onVODSourceFailed` within seconds (the client's shifted-window re-request is the designed recovery); on seekable VOD the reanchor genuinely works.

3. **The advance park cannot wait on an unadvertisable index.** The park releases on a consumer fetch of `head − bufferAheadSegments`, but a sequential append playlist advertises only what this pump's OWN finalize reports have fed it — a release target beyond that frontier is waiting for oneself (field: parked at target=364 while the playlist ended at seg61, freezing it for good). The park is skipped while the target lies beyond the frontier (`sequentialParkWouldSelfDeadlock`, pure); the disk budget stays the resource bound, normal parking resumes as the frontier catches up, and negative startup targets are exempt (they release instantly and are no deadlock).

## Deliberately out of scope

`OutputTimestampSanitizer` keeps latching after a huge jump. movenc latches monotonicity on its own once a wrapped packet is accepted, so a sanitizer re-latch would only convert garbage timestamps into rejected writes until the next muxer rotation — no better — and it would weaken the SSAI seam protection the class exists for. With #368 the sanitizer never sees the leap; without it, containment 2 fails the session cleanly.

## Tests

- `VideoSampleDurationTests`: the cap pinned with the exact wrap values; existing #92 telescoping semantics unchanged under a production-scale cap; at-cap delta still telescopes.
- `SegmentCacheFoldRunTests` (new): the 312-index field run counts, repeats across a producer restart increment (the #358 signal), stored indices stay exempt, narrow runs unchanged.
- `SequentialParkFrontierTests` (new): the field deadlock (target=364 vs frontier=61), at/below-frontier parks normally, negative startup target exempt.
- Full suite green on macOS: 1835 Swift Testing + 504 XCTest, 0 failures.

## Test plan

- Device: Apple TV 4K, tvOS 26.6, SRF Xtream timeshift (H.264 720p50 AC-3, field-coded, open-GOP) — the trace this containment is written against; verified through the Syravo client after the next release is pinned.

Applies cleanly with or without the #368 PR; the combination is integration-tested locally (full suite green on the merged tree).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,20 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Fixed
+
+- A timestamp leap that escapes the timeline rebase no longer turns a VOD session into a
+  long-lived zombie (#369). Three containment gaps, one field trace: the look-behind sample
+  duration is now capped at the discontinuity threshold instead of handing movenc the wrap
+  itself as a duration (device: 8226410192 ticks, rejected as invalid, packet silently lost —
+  the write rc is now logged on first failure too); discontinuity-scale fold runs now reach
+  the fold counters instead of being discarded above 64 indices, so the #358 recovery arms
+  actually arm for exactly the folds most certain to trigger them; and the advance-path
+  backpressure park skips a release target beyond the sequential playlist's advertisable
+  frontier, which only this pump's own finalize reports can move — parking on it was waiting
+  for oneself. Deliberately unchanged: `OutputTimestampSanitizer` keeps latching, because
+  movenc latches monotonicity on its own once a wrapped packet is accepted, and a sanitizer
+  reset would only convert garbage timestamps into rejected writes.
 
 ## [6.25.1] - 2026-08-13
 
--- a/Sources/AetherEngine/Video/HLSSegmentProducer.swift
+++ b/Sources/AetherEngine/Video/HLSSegmentProducer.swift
@@ -205,6 +205,15 @@ final class HLSSegmentProducer: @unchecked Sendable {
     private var seqNextReportIndex: Int? = nil
     private var seqReadyReports: [Int: Double] = [:]
 
+    /// #369: highest sequential index the append playlist can currently advertise (only entries
+    /// with a real duration get a URI). Pump-thread only; read by the advance park to detect a
+    /// release target beyond the advertisable frontier.
+    private var seqHighestAdvertisedIndex = Int.min
+
+    /// #369: one-shot latches for the containment logs. Pump-thread only.
+    private var loggedVideoWriteFailure = false
+    private var loggedSequentialParkSkip = false
+
     /// Order-preserving funnel for sequential finalize reports.
     private func emitSequentialReport(index: Int, duration: Double) {
         let base = seqNextReportIndex ?? index
@@ -213,9 +222,19 @@ final class HLSSegmentProducer: @unchecked Sendable {
         var next = base
         while let d = seqReadyReports.removeValue(forKey: next) {
             onSequentialSegmentFinalized?(next, d)
+            if d > 0 { seqHighestAdvertisedIndex = next }   // #369: zero-duration holes get no URI
             next += 1
         }
         seqNextReportIndex = next
+    }
+
+    /// #369: the advance park releases on a consumer fetch of `target`, but a sequential append
+    /// playlist can only advertise up to the frontier this pump's OWN finalize reports have fed
+    /// it — parking on an index beyond that is waiting for oneself (field case: a fold-to-tail
+    /// parked at target=364 while the playlist ended at seg61). A negative target releases
+    /// instantly, so it is no deadlock. Pure for the unit test.
+    static func sequentialParkWouldSelfDeadlock(target: Int, highestAdvertised: Int) -> Bool {
+        target >= 0 && target > highestAdvertised
     }
 
     /// Forward discontinuity threshold. Distinct from NOPTS-dts repair (+1 tick scale); only fires on genuine multi-second leaps.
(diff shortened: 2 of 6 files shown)


superuser404notfound:
Merged, thanks for the trace-anchored writeup, the three containments sit where the field log points and the pinned wrap values made the duration cap reviewable in one read. Full suite green on the merge result (1838 Swift Testing tests in 268 suites, 0 failures); the CHANGELOG conflict against the #368 follow-up was resolved by keeping both Unreleased entries.

Two follow-ups pushed on top (71b27b1b, merged in 833e1522), both in the same spirit as the PR:

**The cap now also covers the duration the container DECLARES.** `resolveVideoSampleDuration` capped the inferred delta only, and fell through to `existingDuration > 0 ? existingDuration : fallback` untouched. That branch is not a rare path here: it is exactly what runs when no forward delta exists, which is the EOF tail of the same wrapped stream, and movenc rejects a sample on the number, not on where the number came from. Pinned with the same 8226410192 ticks arriving as a declared duration.

**The skipped park now hands its wedge detection to the disk park.** This one is worth a note on the reasoning in section 3. The advance park is what catches a consumer that stopped advancing, and the #207 disk park documents in so many words that it needs no wedge breaker because "the extras eviction that follows the advancing playhead releases it". Skipping the advance park moves the pump straight into that assumption with the consumer possibly frozen: it races to the retention budget (2 GiB, so ~an hour of a 5 Mbps timeshift) and then holds in an unbreakable park for good. The skip is right, so the fix is to carry the detector across rather than to park: `awaitPrefetchDiskBudgetRelease(detectWedge:)` arms the same `BackpressureWedgeDetector` when the advance park was skipped, and its one-second poll matches the cadence the detector counts in. A trip ends the pump onto the existing re-anchor surface, which a sequential origin refuses into `onVODSourceFailed`, i.e. the failure the PR wants, in seconds instead of never.

One correction to the rationale, measured rather than argued, because it changes which path does the work:

- On a **sequential** origin the fold counters cannot reach the #358 arms. `HLSLocalServer` skips zero-duration entries on an EVENT playlist (`if provider.playlistType == .event, dur <= 0 { continue }`), so a folded index never gets a URI and AVPlayer never requests it, which is the trigger `handleTargetChange`'s arm needs. The other arm (`handleBackpressureWedge` reading `consumerTargetFold`) runs only after a #65 wedge break, and section 3 now skips that park on exactly this path. So the "escalation reaches `requestRestart` within seconds" is not what runs on a sequential origin today; the follow-up above is what puts a clean failure back there.
- Counting wide runs is still right, and the win lands on **plain VOD**, where the playlist is the static plan and every folded index IS advertised. That is the source shape #358 was written against (40 s GOP), and the 64-index guard silently disarmed the recovery for the widest folds there. Also worth recording: the guard's stated premise ("a jump this wide is a restart or a seek") could not happen anyway, because `advanceMuxer` only ever fires inside one pump; a restart builds a new muxer at its own base and records no fold run at all.

The issue stays open for the device retest; the release will bundle this with #368 and, if it lands, #370.

Issue #376, opened 2026-08-15
Machine-readable companion to `.error(String)` so a host can classify a failure


superuser404notfound:
`PlaybackState.error` carries a human sentence and nothing else. On the native paths that sentence is `AVPlayerItem.error.localizedDescription`, forwarded verbatim, so it is localized into the device language and the `NSError` domain and code behind it are gone by the time a host sees it. The engine-authored messages are English-stable, but both halves arrive through the same publisher, so nothing distinguishes them except the text.

A host that buckets its failures (analytics, a fallback to a second engine) therefore has to substring-match a localized string, which puts every non-English device into "other". Raised on #374 by a downstream host building exactly that classifier, whose telemetry contract additionally forbids raw player error strings leaving the device.

**Add an additive, machine-readable companion published alongside the state:**

- a stable engine-authored kind (`PlaybackErrorKind`, string-backed so a host can ship it as a token, and open so a new kind cannot break an exhaustive switch),
- the underlying `NSError` domain and code where a Foundation / AVFoundation failure is involved,
- the same message the state carries.

Assigned before `state`, so a `$state` sink can read it synchronously, and cleared by the state's own transition away from `.error` so the two cannot drift.

A numeric code is origin-free by construction, which also gives a host that cannot ship the string something it can ship.

superuser404notfound:
Shipped in [6.26.0](https://github.com/superuser404notfound/AetherEngine/releases/tag/6.26.0).

PR #379, opened 2026-08-16
fix(io): a refused source fails with its HTTP status, not as invalid data (#378)

### LINKED ISSUE #378: A refused source (HTTP 403 on every GET) reaches the host as "Invalid data found when processing input" and disables the tail prefetch for the origin
## What happened

An Xtream panel behind a portal's 302 answered every GET for a VOD `.mkv` with **403** for about twenty seconds (its one-connection slot was still held), while a HEAD came back 200 with `Content-Length: 0`. The reader's log, in order:

```
[AVIOReader] pump tail prefetch issued bytes=-65536
[AVIOReader] pump conn start gen=1 offset=0 len=32MB
[AVIOReader] pump tail prefetch rejected after 762ms: status=403 (no suffix range support); not asking this origin again this session
[AVIOReader] pump gen=1 rejected response status=403 at offset 0
[AVIOReader] Data connection resolved no size, falling back to probe
[AVIOReader] Range probe (bytes=0-) didn't yield a size
[AVIOReader] Cached resolved URL host=…
[AVIOReader] File size: 0 bytes (HEAD fallback)
[AVIOReader] Range probe (bytes=0-1) didn't yield a size
[AVIOReader] no probe resolved a size, streaming mode (forward-only)
[AVIOReader] Streaming started: 1325105.mkv
[AVIOReader] Streaming ended
[AetherEngine] probe failed (Demuxer: open failed (Invalid data found when processing input (-1094995529))); proceeding without criteria
…
state = .error("Failed to load: HLSVideoEngine: open failed (Demuxer: open failed (Invalid data found when processing input (-1094995529))))
errorInfo = sourceOpenFailed, underlyingDomain = "AetherEngine.HLSVideoEngine.HLSVideoEngineError", underlyingCode = 0
```

Same sequence on all four opens (probe + session, twice through the host's retry), then the host showed its "unsupported format / invalid video data" copy for what was a refused connection.

## Why

1. **The forward-only streaming GET never looks at the status.** `StreamingDelegate.didReceive` calls `completionHandler(.allow)` unconditionally (AVIOReader.swift:3313 on 6.26.0), so the 403 error page is appended to `streamBuffer` and `readStreaming` hands it to FFmpeg, which reports `AVERROR_INVALIDDATA` for what was a refusal. The 403 the data connection had already been refused with (`persistentReceivedResponse`) is kept only as the Int `connStatus` and then read by `open()` as "resolved no size" (:691), so the status is lost before the probe ladder even starts.
2. **#376's classification cannot carry it.** `PlaybackErrorKind` has no HTTP-status kind, and the underlying error is a Swift enum, so `underlyingCode` is 0. A host cannot tell "the origin refused" (retry later, another device holds the slot, subscription) from "the file is corrupt".
3. **One 403 disables the tail prefetch for the origin for the process lifetime.** `TailPrefetchDelegate` takes ANY non-206 as `status=… (no suffix range support)` and `SuffixRangeSupport` latches it (#281 / 6.5.5). A 403 during a connection-cap window, a 429 or a 5xx says nothing about suffix ranges, but the origin is never asked again — on an MKV that is the tail read every later open pays for.
4. After the 403 at byte 0 the reader still issues the HEAD and `bytes=0-1` probes, which the same origin answers the same way; on a one-slot panel those are three more connections against a slot that is already full.

## Proposal (PR to follow)

- `StreamingDelegate` hangs up at the header on anything but 200/206 and records the status; `open()` fails typed with a new `AVIOReaderError.httpStatus(Int)` (the AE#140/AE#154 shape), which `HLSVideoEngine.openFailure(from:)` passes through like the HLS classifications, and `load()` publishes as a new `PlaybackErrorKind.sourceRefused` with `underlyingCode` = the status (additive; the struct is string-backed for exactly this).
- After a 401/403/404/410 on the data connection at byte 0, skip the HEAD / `bytes=0-1` probes; still try the one unranged GET (an origin that refuses `Range` but serves a plain GET plays forward-only, one that refuses both fails with its status). 5xx and 429/503/509 keep the ladder.
- The suffix-range denial latches only on the origin's answer to the range FORM (a 200 that ignored it, a 416); 401/403/404/429/5xx neither latch nor count as a transport strike.

Tested against a scripted loopback origin (HEAD 200/0, GET 403 with an error page); on-device: Apple TV 4K (tvOS 26), Xtream VOD MKV behind a 302, `max_connections=1` panel.

tschuegy:
Fixes #378.

## What

An origin that answers the source request with an HTTP status instead of media (a 401/403 refusal, a 404, a 5xx) reached the host as `sourceOpenFailed` carrying FFmpeg's "Invalid data found when processing input", indistinguishable from a corrupt file — and one 403 on the tail prefetch switched the prefetch off for the origin for the rest of the process.

- **Streaming GET checks the status.** `StreamingDelegate` hangs up at the response header on anything but a 200/206 and records the status; `open()` fails typed with a new `AVIOReaderError.httpStatus(Int)` (the AE#140/AE#154 shape). `HLSVideoEngine.openFailure(from:)` passes it through like the HLS classifications, and `load()` publishes it as the new **`PlaybackErrorKind.sourceRefused`** with `underlyingCode` = the status (additive; the struct is string-backed for exactly this). `underlyingDomain` stays nil: the sentence is the engine's.
- **No probe ladder after a refusal at byte 0.** After a 401/403/404/410 on the open-time data connection the HEAD and `bytes=0-1` probes are skipped; the one request still made is the unranged GET, so an origin that refuses `Range` but serves a plain GET plays forward-only, and one that refuses both fails with its status. 5xx and 429/503/509 keep the ladder (and their seekability).
- **The suffix-range latch takes only the origin's answer to the range form.** A 200 that ignored `bytes=-n` or a 416 still latches (#281 stays); a 401/403/404/429/5xx neither latches nor counts as a transport strike.

A `readStreaming` safety net returns EIO (not EOF) if the refusing header arrives after `open()`'s 15 s budget — untyped, but never "invalid data" for a refusal.

## Why this shape

- `sourceRefused` rather than an HTTP domain on `sourceOpenFailed`: the #376 docs say a non-nil `underlyingDomain` marks OS-localized text, and `.vodSourceFailed` is the precedent for "the kind says what `underlyingCode` means".
- The probe skip is limited to `isResolvedExpiryStatus` (401/403/404/410), the predicate the file already treats as "not transient"; a HEAD from the same client is answered the same way, and a size learned from it would only re-issue the refused range on the persistent path, which then dies after one retry with the status lost.
- Not skipping the second open (probe → session reopen): that is the established failure matrix, and it costs one round on a refusing origin.

## Test plan

- `swift test`: 1874 Swift Testing tests + XCTest, all green on macOS 26 / Xcode 26.
- New `Issue378SourceRefusedTests` against a scripted loopback origin: HEAD 200/0 + GET 403 with an error page → `open()` throws `httpStatus(403)`, `cumulativeBytesFetched == 0`, exactly one suffix request, one unranged GET, no HEAD / `bytes=0-1`; the latch stays clear and a second open asks the suffix form again; a 416 still latches; a Range-refusing origin plays forward-only without probes; `sequentialOnly` is typed too; `Demuxer.open` throws the reader error, not `DemuxerError`; `openFailure(from:)` passes it through; `publishLoadFailure` publishes `sourceRefused`/403 vs `sourceOpenFailed`.
- `PublicAPIDocumentationTests` / `DocumentedConstantsTests` green; `docs/api.md` names the kind and the contract, `docs/architecture.md` the reader behaviour.
- Field: Apple TV 4K, tvOS 26, Xtream VOD MKV behind a portal 302, `max_connections=1` panel (the reported origin) — the host now receives `sourceRefused` / 403 during the slot-held window and plays once the slot frees; will confirm on-device after the release lands in the host's pin.


superuser404notfound:
Merged as abbcd90b. Thank you for this one: the diagnosis was exact down to the line, and the shape you picked (typed reader error passed through \`openFailure(from:)\`, additive kind on a string-backed struct) is the one the file already uses, so there was nothing to argue about. The scripted-origin tests are what made the probe-skip claim checkable rather than plausible.

The branch was written against 6.26.0 and main has since taken #377 (6.28.0), so two things were pulled in afterwards, in 4bb63c8e:

- **A metered open publishes `sourceRateLimited`, not `sourceRefused`.** #377 split 429/503/509 off precisely so a host waits instead of handing the source to a second player that the same origin will meter in turn. Typing the open made that split reachable at open time too, and publishing all of it as `sourceRefused` would hand a host the one verdict the contract exists to prevent. The status stays in `underlyingCode` either way; every other status keeps `sourceRefused` exactly as you built it.
- **The streaming pump charges a rate limit against the origin request budget.** Not your defect, a hole in the #377 accounting that your change made visible: a metering origin is charged wherever a status is first read, and the pump was the one path that read one without charging it. On a sequential origin that GET is the session's only request, so its 429 was seen by nobody.

The comment there first named a raw live source as the case, and the test said otherwise: with prefetch on, a live open goes through the persistent connection, which charges the budget already. The sequential origin is the real gap.

Conflicts were only in CHANGELOG.md and docs/api.md; the sources merged clean. 1901 tests green, CI green on main. Release still to be cut, and the issue stays open for the on-device retest you offered.

Issue #69, opened 2026-06-26
Persistent AVIOReader reconnect storm on non-faststart / coarsely-interleaved remote MP4 (triggers origin 429, blocks playback)


ghost:
### What happened?

A remote, non-faststart MP4 (~820 MB, trailing moov, coarsely interleaved) never starts playing. After the header and trailing-moov reads, the persistent AVIO reader tears down and reopens its HTTP connection on nearly every demuxer seek (Persistent conn gen=N ended with error: cancelled); the origin eventually returns 429 and the read loop spins without making forward progress. A faststart file from the same origin plays fine. Expected: read the header/index and begin playback.

### Steps to reproduce

1. Serve a remote MP4 over HTTP(S) that is non-faststart (trailing moov) and coarsely interleaved (track data regions tens of MB apart), so the demuxer must seek between distant file regions during avformat_find_stream_info / index parsing.
2. Load it through the persistent reader path (known Content-Length, prefetch enabled).
3. Watch the [AVIOReader] demux logs: a reconnect-per-seek storm, climbing gen=N, then status=429. Playback never starts.

### AetherEngine version or commit SHA

3.13.2

### Host app

Custom / my own integration

### Platform

macOS

### OS version

27 beta

### Device / chip

m2 / a19 pro 

### Playback path

Native AVPlayer

### Source media (for playback bugs)

MP4 (mov,mp4), 820,727,423 bytes, progressive HTTP(S) download. Non-faststart: moov near EOF (first seek to offset ≈820,001,011). Coarsely interleaved: parser ping-pongs between ≈0–12 KB, ≈21 MB, and the moov region. FFmpeg also logs stream 0, timescale not set during parse. (Codec details via ffprobe or CDN URL  available on request.)

### Error codes / log lines

```text
[demux] [AVIOReader] File size: 820727423 bytes (Range probe)
[demux] [AVIOReader] Persistent conn start gen=1 offset=0
[demux] [AVIOReader] Persistent conn gen=1 ended with error: cancelled
[demux] [AVIOReader] Persistent conn start gen=2 offset=820001011        # trailing moov
[ffmpeg] [mov,mp4,m4a,3gp,3g2,mj2] [warning] stream 0, timescale not set
[demux] [AVIOReader] Persistent conn start gen=3 offset=176
[demux] [AVIOReader] Persistent conn gen=3 ended with error: cancelled
[demux] [AVIOReader] Persistent conn start gen=4 offset=21048761
[...]

PR #381, opened 2026-08-16
fix(io): bound the rate-limited keep-pin grace, then spend one re-resolve (#380)

### LINKED ISSUE #380: A stale pinned redirect target answers 509 forever after a pause-resume; the keep-pin rule has no bound, the streak resets on window-served reads, and the faulted-refill pacing is erased by its own reconnect
### What happened?

Paused a VOD series episode for a few minutes on the couch, pressed play again. Playback resumed
for a few seconds (the resident read-ahead), then picture and sound froze for ~60 s, then the
session healed by itself. No user action in between.

The log tells a clean story: after the resume, every reconnect of the persistent reader against
the **pinned post-redirect URL** was answered **509**, for **20 generations across ~85 s** (gen
8→20, plus the ladder restarts in between). The fresh demuxer the segment layer eventually built
— whose brand-new reader resolves through the **source URL** and follows a fresh 302 — got its
first data after **452 ms** from the same host. The origin was never down; the pinned target's
session was dead.

This is the *other* 509 shape, the one the #307 follow-up's keep-pin rule (`8a8e00b0`, 6.15.2)
wasn't written for. That rule models the lingering slot: the connection being replaced still
occupies the panel's only slot, it frees in seconds, the pin is fine, re-resolving wastes the one
request there is no room for. All still true — for seconds. Here the reader had held **no
connection at all for minutes** (the #310 high-water end: a paused player holds no flow), the
edge host reaped the session server-side, and after that the pinned URL can only ever answer 509.
Only a fresh redirect heals it, and the keep-pin rule has no bound, so the reader never asks for
one.

Two smaller mechanisms turn that policy gap into the full 60 s freeze:

1. **The streaks reset on window-served reads, not on network progress.** The window-serve branch
   of `readPersistent` does `unproductiveReconnects = 0; rateLimitStreak = 0` on every served
   read (`AVIOReader.swift:1406-1407` @ `73354ec8`) — in the same loop iteration that then makes
   the faulted-refill decision. So while the runway drained, every replacement attempt was
   charged `streak=1` (the log shows `replacing a faulted connection … (streak=1 status=509)`
   three times with 4100 KB → 1028 KB → 0 KB of read-ahead left), the bounded give-up was
   unreachable, and after the empty-window ladder finally exhausted at 7, one served byte reset
   the whole ladder and it started over (`… exhausted … (7 consecutive)`, then `(8 consecutive)`,
   then a fresh `streak=1`).

2. **The faulted-refill pacing is erased by the reconnect it authorises.**
   `chargeFaultedRunwayRefill` sets `nextFaultedRefillAt = now + delay`
   (`AVIOReader.swift:1682`), then the caller runs `startPersistentConnection`, which resets it
   to `.distantPast` (`AVIOReader.swift:2057`). The announced "next attempt in 1.0s" is never
   honoured — attempts run on read cadence against the origin that is refusing them.

Because the reader neither fails the read while the window can serve (by design) nor ever
re-resolves, the session only heals once the segment layer runs its full escalation: the 30 s
forward-wait, the march-dead second GET, the 5 s stopWait, a fresh demuxer. That is the ~60 s the
viewer sits through.

### Steps to reproduce

Needs an Xtream-style origin whose stream URL 302s to an edge/load-balancer host that expires
idle sessions (mine: aggregator → panel → 302 → edge, `max_connections=1`).

1. `engine.load()` a progressive-MP4 VOD through such an origin; play a couple of minutes.
2. Pause. The reader ends its connection at high water (#310) — no flow held, correct.
3. Wait 3–5 minutes; the edge reaps the session server-side.
4. Resume. The low-water refill reconnects to the pinned edge URL → 509 → the storm above:
   playback runs a few seconds off the window, then freezes ~60 s until the #169 restart path
   rebuilds a demuxer whose reader resolves through the source.

6.26.0

### Host app

Custom / my own integration

### Platform

tvOS

### OS version

tvOS 26.6

### Device / chip

Apple TV 4K (3rd gen)

### Playback path

Native AVPlayer

MP4 via HTTP progressive, h264 1920x960, 2× AAC audio, 8× ASS subtitle streams, 915 MB,
duration 2485 s. Served by an Xtream-style panel: HTTPS front → 302 → plain-HTTP edge host,
`max_connections=1`, no suffix-range support.

### Error codes / log lines

```text
[AVIOReader] pump replacing a faulted connection at offset 72017621 with 4100KB of read-ahead left (streak=1 status=509, next attempt in 1.0s)
[AVIOReader] pump conn start gen=8 offset=72017621 len=32MB
[AVIOReader] pump gen=8 rejected response status=509 at offset 72017621
[AVIOReader] pump conn gen=8 ended with error: Abgebrochen
[AVIOReader] pump replacing a faulted connection at offset 72017621 with 1028KB of read-ahead left (streak=1 status=509, next attempt in 1.0s)
[AVIOReader] pump replacing a faulted connection at offset 72017621 with 0KB of read-ahead left (streak=1 status=509, next attempt in 1.0s)
[AVIOReader] pump conn ended at offset 72017621 status=509, reconnecting (streak=2 retryAfter=0.0s)
   … streak=3,4,5,6 with real backoff …
[AVIOReader] pump reconnect exhausted at offset 72017621 status=509 (7 consecutive rate-limited)
[AVIOReader] slow read: 44581ms at offset=72013317 detour=0(0ms,0fetch) stallWaits=6(304ms,6signaled) tailWaits=0(0ms) reconnects=6 backoff=44272ms lockWait=0ms connect=3ms staleGenDropped=0b iters=13 unaccounted=1ms gen=9->15
[AVIOReader] pump reconnect exhausted at offset 72017621 status=509 (8 consecutive rate-limited)
[AVIOReader] pump replacing a faulted connection at offset 72017621 with 0KB of read-ahead left (streak=1 status=509, next attempt in 1.0s)
   … gen=16..20 all 509 …
[HLSVideoEngine] seg30: cache miss after 30001ms (cache=25 restarted=false)
[HLSVideoEngine] seg30: #169 forward-wait march dead (front=29 frozen across a full wait); escalating to restart
[HLSVideoEngine] restart at idx=30: old producer wedged in a read past 5s; aborted it and reopened a fresh demuxer
[AVIOReader] pump conn start gen=1 offset=0 len=8MB
[AVIOReader] File size: 915393993 bytes (data connection)
[AVIOReader] Cached resolved URL host=<edge>
[AVIOReader] pump gen=1 first data after 452ms
```

### Anything else

The retryAfter is always 0.0s — this panel sends no Retry-After on its 509s.

I have a fix ready and will open a PR right after this: the rate-limit keep-pin gets a bounded
grace (3 paced attempts, then the pin is dropped for exactly one re-resolve; a 200/206 re-pins),
the streaks reset only when the current generation has delivered data, and the faulted-refill
timestamp survives `startPersistentConnection` (released by first data or a seek). The two-attempt
lingering-slot tests in `ResolvedURLInvalidationTests` stay green — that shape keeps its grace —
and the permanent-509 case keeps the same bounded give-up, now with the re-resolve spent inside
the same seven attempts. Verified against a two-server loopback rig (source mints per-request
lease URLs, CDN 509s dead leases forever): the unfixed reader burns 20 boundary attempts and
fails the read; the fixed one heals in 4 attempts with exactly one source resolve.

tschuegy:
Fixes #380.

## What

A pause of a few minutes, then resume, froze VOD playback for ~60 s on a connection-capped
panel: the pinned post-redirect URL answered 509 to 20 consecutive generations (~85 s) while a
fresh redirect through the source connected in 452 ms. Three coupled defects kept the reader
grinding the dead pin until the segment layer tore the producer down from outside — the full
trace and analysis are in the issue.

## How

- **The keep-pin grace is now bounded.** The rule from the #307 follow-up (`8a8e00b0`) models
  the lingering slot, which frees in seconds — and for that shape nothing changes: the pin
  survives `rateLimitRepinStreak` (3) paced attempts, which is the existing two-attempt tests'
  territory plus margin. A rate-limited streak that reaches the rung is the other shape (a
  pinned edge session that expired during a long idle and refuses forever), and there the pin
  is dropped for exactly one re-resolve through the source; the fresh target's 200/206 re-pins.
  Both ladders (`chargeFaultedRunwayRefill` and the empty-window path) share the rung; the
  hard-error rung (`unproductiveReconnects >= 2`) is unchanged.
- **Window-served reads no longer reset the streaks.** Draining read-ahead is not network
  progress: the reset ran in the same read iteration as the faulted-refill decision, so every
  attempt against the refusing origin was charged streak=1 while the runway drained, the
  bounded give-up and the new rung were unreachable, and one served byte after an exhaustion
  restarted the whole ladder. The streaks now reset only when the current generation has
  delivered data (`connFirstDataSeen`, already winCond-guarded in that branch).
- **The faulted-refill pacing survives the reconnect it authorises.**
  `startPersistentConnection` reset `nextFaultedRefillAt` as part of starting the very attempt
  the charge had just paced, so "next attempt in Ns" never held and attempts ran on read
  cadence. The reset moved to the two events that genuinely end a faulted lineage: first data
  (`appendPersistentData`) and an intentional reposition (`seekReconnect`). The `.distantFuture`
  give-up latch actually latches now, and first data re-arms it for a later, unrelated fault.

## Test plan

- `swift test` on macOS 26: **1867 tests green**.
- New in `ResolvedURLInvalidationTests` (both verified failing on the unfixed base):
  - *"a 509 outliving the keep-pin grace re-resolves once through the source"* — a two-server
    rig where the source mints per-request lease URLs and the CDN 509s dead leases forever.
    Fixed: heals at `rateLimitRepinStreak + 1` boundary attempts with exactly one post-fault
    source resolve. Unfixed base: 7 attempts, zero resolves, read dies at the boundary.
  - *"a draining window does not reset the rate-limit ladder"* — a paced consumer (2 MB/s)
    draining 2 MB of runway against a permanent 509. Fixed: bounded at ≤ 7 boundary attempts,
    every pre-boundary byte served, then a clean −1. Unfixed base: 20 attempts.
- Updated: *"a permanent 509 pays the rate-limit ladder and gives up cleanly"* now expects the
  bounded re-resolve (≥ 1, ≤ 4 source hits) inside the same seven attempts.
- Unchanged and green: the two-attempt lingering-slot cases (503 and 509 keep the pin end to
  end), the hard-500 fallback, `ErrorEndedReconnectTests`, `Issue309SilentTransportDeathTests`
  (including the parked-is-not-stalled regression), the live-window suites, and the detour cap
  unit test.
- Field shape: Apple TV 4K (tvOS 26) against an Xtream aggregator (`max_connections=1`) whose
  edge host expires idle sessions; device retest against that origin planned as soon as a
  release with this lands on the app's pin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,30 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Fixed
+
+- **A rate-limited streak that outlives the lingering-slot grace drops the pinned redirect target
+  for one re-resolve through the source (#380).** 509 has two field shapes. The one the keep-pin
+  rule was built for (#307 follow-up): a connection-capped panel refusing while the slot of the
+  connection being replaced lingers — it frees in seconds and the pin is fine. The one it broke:
+  a resume after minutes of pause, where the reader held no connection (#310) and the pinned edge
+  target's session expired server-side, so it answers 509 forever while a fresh redirect through
+  the source connects on the first try (field trace: 20 generations of 509 across ~85 s at one
+  offset, then a source-resolved reader delivered first data in 452 ms). The pin now survives
+  `rateLimitRepinStreak` (3) paced attempts and is then dropped for exactly one re-resolve; the
+  fresh target's 200/206 re-pins, and a permanently metering origin pays the same bounded
+  give-up as before, with the re-resolve spent inside the same seven attempts.
+- **Window-served reads no longer reset the reconnect streaks (#380).** Draining read-ahead is
+  not network progress: the reset ran in the same read iteration as the faulted-refill decision,
+  so a refused replacement was charged streak=1 for as long as the runway lasted, and neither
+  the re-resolve rung nor the bounded give-up was reachable until the window was empty — and one
+  served byte after an exhaustion restarted the whole ladder. The streaks now reset only when
+  the current generation has delivered data.
+- **The faulted-refill pacing survives the reconnect it authorises (#380).**
+  `startPersistentConnection` reset the next-attempt timestamp the ladder had just set, so
+  "next attempt in Ns" fired as fast as the consumer could read, and the give-up latch was
+  erased by the next reconnect from any path. The timestamp is now released by first data or an
+  intentional reposition — the two events that genuinely end a faulted lineage.
 
 ## [6.26.0] - 2026-08-15
 
--- a/Sources/AetherEngine/Demuxer/AVIOReader.swift
+++ b/Sources/AetherEngine/Demuxer/AVIOReader.swift
@@ -113,7 +113,10 @@ final class AVIOReader: AVIOProvider, @unchecked Sendable {
     /// connection-capped IPTV panel answers while its slot is still occupied by the
     /// connection being replaced — the slot frees in seconds, the pinned redirect target
     /// is fine, and re-resolving through the portal spends the one request there is no
-    /// room for (519ae26e, #307 follow-up).
+    /// room for (519ae26e, #307 follow-up). That grace is bounded, not absolute: a 509
+    /// that outlives `rateLimitRepinStreak` paced attempts is a pinned edge target whose
+    /// session expired (a resume after minutes of pause), and there the pin is dropped
+    /// for one re-resolve through the source.
     static func isRateLimitStatus(_ status: Int) -> Bool {
         return status == 429 || status == 503 || status == 509
     }
@@ -321,6 +324,13 @@ final class AVIOReader: AVIOProvider, @unchecked Sendable {
     // Distinct axis from unproductiveReconnects: NOT reset by seekReconnect, so parse-driven
     // seeks cannot mask a throttled origin into an infinite reconnect loop (AetherEngine#71).
     private static let rateLimitMaxStreak = 6
+    // Rate-limited attempts that keep the pinned redirect target before one attempt through the
+    // source URL is spent on a fresh redirect. Three paced attempts (~7 s of ladder) ride out the
+    // lingering-slot 509 of a connection-capped panel (#307 follow-up: the slot frees in seconds);
+    // a streak that reaches this rung is the other shape — a pinned edge target whose session
+    // expired during a long pause and refuses forever, where only a re-resolve heals. Internal so
+    // the rung is unit-tested without a live origin.
+    static let rateLimitRepinStreak = 3
 
     /// NSCondition guards all persistent-mode fields and serves as the
     /// edge-triggered condition variable for read waits and backpressure.
@@ -1403,8 +1413,16 @@ final class AVIOReader: AVIOProvider, @unchecked Sendable {
                 position = curPosition + Int64(copyNow)
                 totalRead += copyNow
                 trimWindowLocked()
-                unproductiveReconnects = 0      // real progress
-                rateLimitStreak = 0             // real progress clears the 429 give-up streak (#71)
+                // "Real progress" is the CURRENT generation having delivered — draining read-ahead
+                // is not. An unguarded reset here ran in the same iteration as the faulted-refill
+                // decision below, so a connection-capped origin refusing every replacement was
+                // charged streak=1 forever while the runway drained (field trace: a post-pause 509
+                // storm held streak=1 across 4 MB of served reads, and the bounded give-up and the
+                // re-resolve rung were both unreachable until the window hit empty).
+                if connFirstDataSeen {
+                    unproductiveReconnects = 0      // real progress
+                    rateLimitStreak = 0             // real progress clears the 429 give-up streak (#71)
+                }
                 emitNetworkPhase(.flowing)      // recovered: source delivering again (#85)
                 // No flow installed and the consumer has drawn down to low water: request at the
                 // frontier. #220/#310 built this for PLANNED ends (a range delivered in full, a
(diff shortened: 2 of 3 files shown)


superuser404notfound:
Merged as d35d078b. Thank you for this one: a report that arrives with the mechanism, the two smaller defects that turn the policy gap into the full freeze, and a loopback rig that reproduces it is about as good as it gets.

Two things worth adding to the record.

## Independent corroboration from #377, filed the same week

A different reporter, a different origin, 429 rather than 509, and mid-playback rather than post-pause. Their trace after retesting the #377 branch:

```
[AVIOReader] slow read: 32513ms at offset=0 ... reconnects=6 backoff=31968ms ... refusals=9
[AVIOReader] pump conn start gen=8 offset=0 len=32MB
[AVIOReader] Cached resolved URL host=nexus-175.neur.tb-cdn.st
[AVIOReader] pump gen=8 first data after 210ms
```

45 s of refusals against the pin, then 210 ms once a fresh demuxer's reader re-resolved. That is your 452 ms, on someone else's CDN. Their pinned host also changes between sessions, so those are per-session signed targets expiring on their own schedule, which is the shape your grace bound is written for. They reached your issue on their own and asked whether it was their root cause seen from the other end. It reads like it is.

## What I pulled in afterwards (#386)

Reviewing the PR against main turned up that the read loop serves from memory in three places, and your fix covers one of them.

- **The retained head/tail spans (#281)** run FIRST, before every network path, and their own log line says "no reconnect for it". They cleared both streaks. That branch is the reachable one on a metered origin: a refusal hands the origin to the #377 budget merged just before your PR, which caps it to one request and takes the detour out of service, while the spans keep serving. So the parse's return to the head is precisely the "one served byte reset the whole ladder and it started over" you described, and it survived the fix written for it.
- **The detour cache's resident-block hit (#69)** did the same. `DetourServe.served` now carries whether the bytes crossed the network.

Also dropped the `#if DEBUG` from both pin lines. With your bounded grace, dropping the pin is a decision the ladder makes rather than a reaction to an expiry status, and a rung that only fires in the field is worth nothing if the field cannot see it.

`rateLimitRepinStreak` is left at 3 as you set it, and the changed expectation in `permanent509TakesTheRateLimitLadder` stands: a bounded number of re-resolves inside the same seven attempts is the right trade against keeping a pin that can only ever be refused.

Verified on the merge with main (your branch was 12 commits behind, and the #377 budget lands in the same file): 1886 tests, 273 suites, green.

Issue #290, opened 2026-08-03
saneSAR admits small-but-wrong ratios: bound the resulting display aspect, not just the components


kskchaitanya1993:
`saneSAR` bounds each SAR component to 256, which catches the pathological values (the `1088:1` the comment mentions) but still admits small-but-wrong ratios that stretch the picture into an unwatchable strip.

Observed on a live 1080p H.264 IPTV channel that declared SAR `3:1`, then `2:1` on a later run — both well inside the existing gate. `3:1` applied to 1920x1080 yields a 5.33:1 display aspect; `2:1` yields 3.55:1. The picture renders as a horizontally smeared band. The user-visible workaround is switching the aspect mode to Fill, which is a real tell that the ratio itself is wrong rather than the content being unusual.

The magnitude gate can't distinguish these, because the problem isn't the size of the numbers — it's the aspect they produce.

### What we do locally

Bound the *resulting* display aspect rather than the component magnitudes, in the same helper:

```swift
static func saneSAR(_ sar: AVRational) -> AVRational? {
    guard sar.num > 0, sar.den > 0, sar.num <= 256, sar.den <= 256 else { return nil }
    // Reject an implausibly extreme pixel aspect (e.g. a bogus 3:1 that stretches 1920x1080 into a
    // 5760x1080 strip). No real PAR stretches a frame more than ~2.5x on either axis.
    let ratio = Double(sar.num) / Double(sar.den)
    guard ratio <= 2.5, ratio >= 1.0 / 2.5 else { return nil }
    return sar
}
```

Rejecting returns `nil`, which falls back to square pixels — for a stream lying about its SAR that is the correct picture, and for a stream telling the truth it's never reached, since no legitimate PAR is that extreme.

### On the threshold

2.5 is a judgement call and I'd expect you to want a different one. The widest genuine PARs in circulation are the anamorphic DVD/broadcast set — 16:11 (~1.45), 12:11, 64:45, and the 4:3-into-16:9 cases around 1.8 — so 2.5 leaves a lot of headroom above anything real while still rejecting the 3:1 above. Verified on our side that legitimate anamorphic content is unaffected: a 720x576 PAL source at 16:11 still resolves to ~16:9 and renders correctly.

This refines the gate you added in #177 rather than replacing it — both checks are useful, they just catch different failures. Raising as an issue rather than a PR since the bound is a policy choice; happy to send it either way.

---
[...]

PR #385, opened 2026-08-16
Expose the source sample aspect ratio and flag natively rendered subtitle tracks

Rasmusmart57:
Two small additive properties, both from data the engine already reads and then drops.

`sourceVideoWidth` and `sourceVideoHeight` are the coded dimensions, so on anamorphic video they aren't the shape the picture actually is. I use them to work out a natural size for positioning my own overlays and there's nothing to correct with, and AVAsset track properties don't help on the loopback path since the item there is the engine's own HLS rather than the source. The probe already reads the stream, so this is one more field off the same `AVStream`. Stream SAR wins over codecpar SAR when both are set, stays 1 when neither is.

The other one is for the remote-HLS path, where the backend renders WebVTT itself and no cue ever reaches `subtitleCues`. From outside those tracks look identical to any other, so my subtitle picker was offering position and delay controls that can't affect them. Selecting one now goes down a different path entirely, which needs to know which tracks they are. Set only in `RemoteHLSMediaSelection` where the distinction is already known, defaults to false everywhere else.

I've been carrying both as local patches on a vendored copy and reapplying them by hand on every update. Would rather not keep doing that.

`swift build` and `swift test` clean, 1865 tests, same as without the changes.

### DIFF
--- a/Sources/AetherEngine/AetherEngine.swift
+++ b/Sources/AetherEngine/AetherEngine.swift
@@ -1468,6 +1468,10 @@ public final class AetherEngine: ObservableObject {
     /// is parsed. 0 before load or when source has no video (AetherEngine#28). Also available in SourceProbe.
     public private(set) var sourceVideoWidth: Int32 = 0
     public private(set) var sourceVideoHeight: Int32 = 0
+    /// Display-width multiplier for non-square source pixels. Hosts use this with
+    /// the coded dimensions when positioning their own overlays over anamorphic
+    /// video; the coded size alone is not the shape the picture ends up.
+    public private(set) var sourceVideoPixelAspectRatio: Double = 1
 
     /// MKV font attachments from the probe. Hosts write these to disk for ASS renderer font config (AetherEngine#30).
     /// Not @Published and not in SourceProbe: payloads are 10-30 MB and only playback hosts need them.
@@ -2789,6 +2793,7 @@ public final class AetherEngine: ObservableObject {
         sourceVideoBitrate = 0
         sourceVideoWidth = 0
         sourceVideoHeight = 0
+        sourceVideoPixelAspectRatio = 1
 
         // #114: guarantee the AVAudioSession category is declared (off-main, from init) before any branch
         // below can activate the session: AVKit on the native/remote-HLS paths, activateRendererAudioSession()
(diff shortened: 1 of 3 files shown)


superuser404notfound:
Merged, thanks for both of these, and for carrying them as local patches long enough to know they were worth upstreaming.

The SAR half turned out to sit on top of a defect, which is why it took a moment. Checking what a host would actually lay out against, I cut two fixtures from one 720x576 source, one declaring 64:45 in the H.264 VUI and one declaring it only in the container, and looked at the `pasp` box in the served `init.mp4` next to the software path's display size:

| fixture | software display size | loopback `pasp` | thumbnail |
| --- | --- | --- | --- |
| VUI 64:45 | 1024x576 | 64:45 | 320x180 |
| container 64:45 | 720x576 | 1:1 | 320x256 |

So on the loopback path the second file was not merely missing a number for the overlay, the picture itself was squeezed: `mov_write_pasp_tag` reads the output codecpar, a container-declared ratio reaches `AVStream` alone (matroskadec.c, mov.c), and the muxer's `avcodec_parameters_copy` therefore carried square pixels into the sample entry. The software path and the thumbnail extractor got it wrong for a related reason: a square bitstream ratio was read as a declaration and ended the resolution one axis above the container's.

Fixed in a347858f: the declared ratio is resolved once, in `PixelAspectPolicy`, and the loopback writes that into the `pasp` it serves, so all three agree with what `av_guess_sample_aspect_ratio` returns for the same file. Both fixtures now present at 1024x576 on every path.

`sourceVideoPixelAspectRatio` reads through that same policy now, which changes two things worth knowing about for the overlay use case. It stays 1 where the declared ratio is one the engine refuses (a live channel claiming 3:1 on 1080p, #290), so it can no longer hand out a shape the picture does not have, and it is the same ratio the fMP4 carries, so `nativePlayerLayer.videoRect` and this property agree once the item is laid out. Both are documented in `docs/api.md` now, which is also what the public-API documentation test wanted before the build could go green.

`isNativelyRenderedSubtitle` went in as written and is documented with the rest of `TrackInfo`.

If the anamorphic sources on that side were part of what prompted this, a look at whether they now present at the corrected width would be welcome, especially any that carry the ratio in an MP4 `pasp`.

superuser404notfound:
Shipped in [6.30.0](https://github.com/superuser404notfound/AetherEngine/releases/tag/6.30.0), both properties and the pixel-aspect fix behind them. Both downstream apps are on the new pin already.

PR #394, opened 2026-08-17
fix(audio): publish a rebuffer from the bare-AVPlayer audio host

tschuegy:
## Summary

`AudioAVPlayerHost` published no buffering axis, so on the audio-only path a starved progressive stream sat in `waitingToPlayAtSpecifiedRate` while `playbackPhase` stayed `.playing` with a frozen clock — although architecture.md said a reconnect on this path "reads as `.rebuffering`". This makes that sentence true: the host folds AVPlayer's own signals into a rebuffer flag the engine wires into `isBuffering`.

## What changed

- **`AudioAVPlayerHost.isRebuffering`** (internal, `@Published`): folded by the pure `rebufferingFlag(hasStartedPlaying:status:stalledSinceLastPlaying:)` from `timeControlStatus == .waitingToPlayAtSpecifiedRate` and `AVPlayerItemPlaybackStalled`. A wait counts only once the item has played (the pre-roll of a fresh track is startup, not a rebuffer), never while the player is paused, and a stall notification arriving ahead of the status flip is latched until the next `.playing` rather than lost. Reset per `load()` / `stop()`.
- **Engine wiring** in `loadAudioNative`: `host.$isRebuffering` → `isBuffering` under the existing `state == .playing` gate (same shape as the native video sinks). Transport reconciliation stays off on this path — the transient background `.paused` that once mis-latched Now-Playing is untouched; only the buffering axis moves, so `playbackPhase` reads `.rebuffering` and never flips `state`.
- Item error-log entries and stall notifications are logged under `sw.playback`, because this path has no demuxer/reader logs and a field failure otherwise leaves no trace.
- Docs: `architecture.md` (phase section + file tree), `api.md` (`$playbackPhase` row), `CHANGELOG.md` under `[Unreleased]`.

## Test plan

- `swift test` on macOS 26: **1931 tests green**, including the new `AudioAVPlayerHostRebufferingTests` (startup wait ≠ rebuffer, wait after first play = rebuffer, stall latch covers notification-before-status, paused never rebuffers).
- Device / OS: Apple TV 4K (3rd gen), tvOS 26.6, host app Syravo.
- Source media: internet radio, MP3 192 kbps over HTTPS/ICY (Radio Argovia via QuantumCast, `Accept-Ranges: none`, `icy-metaint: 8192`) → `audio dispatch: codec=86017 -> AVPlayer`.
- Result on the unfixed engine (6.30.0): pausing the host's `AVPlayer` behind the engine's back left `playbackPhase == .playing` with a frozen clock; the host app had to time the clock itself (8 s watchdog) to recover. Session/route logs of that run: HDMI 2ch @ 44.1 kHz, `automaticallyWaitsToMinimizeStalling` on, no error-log entries. Device retest against the same station with a release carrying this change is planned as soon as the app's pin moves.

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Commit messages follow Conventional Commits (`fix(audio): …`)
- [x] Public API changes are intentional and documented (none added; `isRebuffering` is internal, the observable surface is the existing `isBuffering` / `playbackPhase`)


superuser404notfound:
Merged, and hardened on top in b424a4d9. Full suite green here too (1932 tests in 279 suites on macOS 26).

The fold sits in the right place: it is the axis architecture.md has been promising, it never touches `state`, and the sink mirrors the native one in `loadRemoteHLS` including the `.error` bail. Four things around it needed a second pass:

1. **A starve that begins inside a seek window was dropped.** The engine writes `isBuffering` only on the host's rebuffer edges and gates it on `state == .playing`. During `seek()` the state is `.seeking`, so the edge writes `false`, and the seek finalize on this path just sets `state = .playing` again without re-reading anything. A seek into an unbuffered span therefore landed as a frozen `.playing`, which is the shape the axis exists to end. The finalize now re-reads `host.isRebuffering` against the state it just settled (the native path already does that in `reconcileNativeSeekTransport`).
2. **`api.md` overstated the reach.** `.rebuffering` is published on the AVPlayer-backed paths only (`loadNative`, `loadRemoteHLS`, and now `loadAudioNative`). `loadSoftware` and `loadAudio` never write `isBuffering` at all and report a starving source as `.stalled` instead.
3. **architecture.md dated the gap to 6.30.0.** 6.30.1 shipped without the axis too.
4. **The error-log observer read `self.playerItem`** after the actor hop, which after a `load()` swap describes a different item. It reads the item off the notification now, and since nothing in that block touches the host, the hop is gone.

One note on the evidence, and one question. The device run in the test plan pauses the AVPlayer behind the engine's back, and a paused player reports `timeControlStatus == .paused`, which the new fold deliberately reads as "not rebuffering". So that measurement demonstrates a different silent shape rather than exercising the new flag: a sustained external pause, which this path cannot reconcile at all because feeding `timeControlStatus` back mis-latches Now-Playing. The retest against the real station is the one that decides whether a dying ICY origin presents as `waitingToPlayAtSpecifiedRate`, as a stall notification, or as neither, and the two new `sw.playback` lines should make that visible either way.

Which leaves the pause shape open, and it is worth its own issue if the 8 s clock watchdog in the app is covering a real field failure rather than the simulation. A frozen clock under a `.playing` engine is a different lever from transport reconciliation and would not have to touch `state`, so there is room for a signal there.

Thanks for the fix, the tests, and the write-up.

superuser404notfound:
Released in [6.30.2](https://github.com/superuser404notfound/AetherEngine/releases/tag/6.30.2), tag at `2937693`, and both consumers are pinned to it.

The retest that matters is still the one against the real station: whether a dying ICY origin presents as `waitingToPlayAtSpecifiedRate`, as a stall notification, or as neither. The two new `sw.playback` lines should say which, and the item error log is the one that names the failing fetch.

Issue #407, opened 2026-08-23
VC-1 (software path) playback judder due to missing frame PTS on I/P pictures


classicjazz:
### What happened?

VC-1 Advanced Profile content played through AetherEngine's software decode path judders, reproducibly, on Apple TV 4K (via Sodalite/Jellyfin or Moonfin/Jellyfin) and does not judder in mpv or other FFmpeg-based players on the same files on AppleTV. The interlace/deinterlace subsystem, container transport, and file bitrate have all been ruled out. The most likely cause is that AetherEngine's software decoder reads AVFrame.pts directly rather than FFmpeg's best_effort_timestamp, and AVFrame.pts is unset (AV_NOPTS_VALUE) on every I and P frame in affected files — only B-frames carry a valid raw pts. AVSampleBufferDisplayLayer, which the software path uses for presentation, has no independent pacing/cadence model, so it presents strictly on whatever CMTime it's given — any inaccurate fallback timestamp for the 60% of frames lacking raw PTS becomes visible, periodic judder.

### Steps to reproduce

Play a VC-1 file in Sodalite
It will exhibit significant (unwatchable) video judder 

### AetherEngine version or commit SHA

as embedded in Sodalite 1.0.0 (20)

### Host app

Sodalite

### Platform

tvOS

### OS version

tvos 26

### Device / chip

Apple TV 4k 3g

### Playback path

Software (dav1d) host

### Source media (for playback bugs)

vc1 (Advanced) (WVC1 / 0x31435657), yuv420p(progressive), 1920x1080, 23.98 fps, 23.98 tbr, 1k tbn, Matroska container, TimecodeScale default (ms granularity)

[ffprobe.txt](https://github.com/user-attachments/files/31355933/ffprobe.txt)

### Error codes / log lines

```text

```

### Anything else

From Claude based on extensive analysis of video files, logs, and testing across devices and media types. I also have separate analysis of why this is an issue with AetherEngine and not mpv and other ffmpeg based media players.

-----

## Environment

- Repo: `github.com/superuser404notfound/AetherEngine`, HEAD examined `4219e140` (2026-08-22), release tag `6.34.1`
- Clients: Sodalite and AetherPlayer both pin `f4bebc0fd3c66f656ef18d2c7c0b3243df3ac86d` = tag `6.34.1` (current release, not stale)
- Playback target: Apple TV 4K
[...]

PR #411, opened 2026-08-24
fix(video): recover H.264 with missing composition offsets

### LINKED ISSUE #409: Progressive H.264 MP4 without CTTS judders on the native path
### What happened?

On a progressive H.264 MP4 VOD with reordered pictures but no composition-time table, the native stream-copy path produces continuous forward/backward judder from the first frame. Seeking is not required to trigger it. The same source plays smoothly in another FFmpeg-based player.

The container presents every sampled video packet with `PTS == DTS` even though `codecpar.video_delay == 1`. A bounded libavcodec decode of the same head sample shows the contradiction directly: raw decoded `AVFrame.pts` regresses 16 times while `best_effort_timestamp` advances 58 times with zero regressions. Because the PTS values are present rather than `AV_NOPTS_VALUE`, the existing `+genpts` open flag and the #407 fallback do not repair them. Stream-copying those packet timestamps into fMP4 preserves decode order as presentation order, so AVPlayer displays future reference pictures before the B pictures that precede them.

Routing only this confirmed shape through the software decoder and scheduling decoded frames on `best_effort_timestamp` removes the judder.

### Steps to reproduce

1. Open a seekable MP4 containing progressive H.264 with B-frame reordering, but whose sample table has no `ctts` composition offsets.
2. Play through the normal native VOD path.
3. Observe continuous small forward/backward motion from playback start.
4. Inspect the first 64 video packets and decoded frames. The reporting source produced: `valid_pairs=64`, `PTS==DTS` for all pairs, `nonkey=62`, `decoded=59`, `best_effort_advances=58`, `best_effort_regressions=0`, `raw_pts_regressions=16`.

A tiny 96x64 H.264 Main, 30000/1001 fps, 3-B-frame synthetic fixture and its healthy `ctts` twin are ready for a regression test in the accompanying PR.

6.39.0 / `f70d4aee595f0947ef2354fbae01c76b41a7857b`

### Host app

Custom integration

### Platform

tvOS

### OS version

tvOS 26.6 beta (23L773)

### Device / chip

Apple TV 4K (3rd generation)

### Playback path

Native AVPlayer

MP4 / progressive H.264 with B-frame reordering / SDR. The source has no usable composition offsets: sampled packet PTS and DTS are equal despite `video_delay=1`. Audio is not involved in the failure. The synthetic regression fixture is MP4 / H.264 Main / 96x64 / 30000/1001 fps / no audio / SDR.

### Error codes / log lines

No AVPlayer or VideoToolbox error is emitted. The diagnostic signature is:

```text
video_packets=64 valid_pairs=64 nonkey=62 decoded_frames=59
best_advances=58 best_regressions=0 raw_regressions=16
raw_best_differences=16
```

### Anything else

This is adjacent to #407 but not the same case. #407 repairs only `AVFrame.pts == AV_NOPTS_VALUE`; here the raw PTS is always present, but it is the wrong decode-order axis. A focused engine patch with bounded fail-closed detection, automatic per-session software routing, and good/bad MP4 fixtures is being prepared.

orut34iop:
## Summary

Recover seekable progressive H.264 MP4 VOD whose sample table omitted composition offsets despite B-frame reordering. A bounded, fail-closed engine probe confirms the malformed timestamp signature, routes only that session through software decode, and schedules frames on libavcodec's monotonic `best_effort_timestamp` presentation axis.

Fixes #409

## What changed

- Add a bounded MP4/H.264 head probe that exits healthy on the first real `PTS != DTS` pair and requires sufficient equal-timestamp, non-key, decoded-frame, raw-PTS-regression, and monotonic best-effort evidence before selecting repair.
- Arm the existing AVIO read deadline, bound the rewind, require rewind after every read attempt (including EOF/error), and discard or fail rather than hand a sampled demuxer to playback at an unknown position.
- Pair the confirmed software-route decision with an internal per-session best-effort timestamp policy, applied immediately after `avcodec_receive_frame` so captions, deinterlacing, seek filtering, and rendering share one repaired clock.
- Preserve the existing #407 `AV_NOPTS_VALUE` fallback and preserve the confirmed route/policy across audio switches and seekable-custom-source background reloads; reset it for every new public load.
- Add healthy/malformed synthetic MP4 integration fixtures, threshold/budget/overflow/rewind/routing/lifecycle tests, format documentation, and a changelog entry. No public API changes.

This is adjacent to #407 but covers a different failure: here every raw PTS is present, so the #407 NOPTS fallback correctly does nothing; the present raw axis itself regresses.

## Test plan

- Device / OS: Apple TV 4K (3rd generation), tvOS 26.6 beta (23L773).
- Source media: seekable MP4 / progressive H.264 with B-frame reordering / SDR / audio not involved; all 64 sampled packet pairs had `PTS == DTS`, while decoded raw PTS regressed 16 times and best-effort PTS advanced 58 times with zero regressions.
- Result: before the patch, continuous forward/backward judder began on the first frame and remained after seeking; with the repaired engine candidate, normal playback and post-seek playback were smooth. A healthy comparison MP4 remained smooth.
- Synthetic coverage: byte-identical H.264 Main elementary streams in 96x64, 30000/1001 fps, 66-frame MP4 twins; the healthy twin carries `ctts`, while the malformed twin was stream-copied with `setts=pts=DTS` and has no composition offsets.
- `swift build` — passed.
- `swift test` — 2,001 tests in 287 suites passed.
- `swift build --build-tests --package-path Probes/TransportProbe` — passed.
- `xcodebuild build -scheme AetherEngine -destination 'generic/platform=tvOS Simulator' ...` — passed.
- TransportProbe tvOS Simulator device harness `build-for-testing` — passed.
- Local Xcode has only the tvOS 26.5 platform installed, so iOS Simulator and visionOS Simulator destination selection could not run locally; the corresponding upstream CI jobs cover both SDKs.

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Public API changes are intentional and documented (no public API changes)


superuser404notfound:
Thanks for this, and especially for the twin fixtures. One encode muxed twice with the composition offsets removed from one is exactly the asset this needed, and I have kept that pair as the regression test for the fix.

The diagnosis holds. I reproduced it on the Apple decoder rather than only in libavcodec, by decoding both twins through `AVAssetReader` and matching each decoded frame's content against the healthy twin:

| | healthy twin | ctts removed |
|---|---|---|
| pictures shown at a time belonging to a different picture | 0 of 66 | **45 of 66** |
| backward steps in the content order presented | 0 | **30** |

What I could not keep is the remedy. Measured on the same pair, `best_effort_timestamp` resolves to `pkt_dts` of the packet being fed, so it is the decode ladder shifted by the reorder delay rather than the presentation axis:

- **+2002 ticks (2 frames, 67 ms) on 63 of 66 pictures**, permanently. `has_b_frames=2` is the x264 default, so that is what a repaired file gets in the common case; at a 3-frame delay it is 100 ms of lip sync error.
- the first interval is **3 frame durations** instead of 1, so the opening picture is held three times too long.
- at EOF the flush regresses: `..., 65065, 59059, 65065`. That is a backward step of 4004 ticks plus a duplicate, which is the artifact class the patch sets out to remove. The `bestEffortRegressions == 0` gate is measured over the first 64 packets, so it structurally cannot see that tail.

It also trades hardware decode away for what is a container metadata defect.

The information the container dropped is still in the bitstream, so #413 rebuilds it instead of routing around it. libavcodec's H.264 parser reads each access unit's picture order count without decoding a pixel, and it accepts MP4's length-prefixed AVCC payload directly, so the repair costs one slice header parse per video packet and no Annex-B conversion. The packet is then rewritten to what the muxer should have written: presentation by display rank, decode pulled back by the reorder delay, which is also what keeps `PTS >= DTS`. Across three fixture pairs, 432 packets, both edit-list shapes and seven IDR boundaries, every repaired packet matches its healthy twin's PTS and DTS exactly, and the served loopback output presents all 301 frames at the same times the healthy twin's output does. The source keeps the native path and hardware decode.

Two things from this patch went straight into that one: the fail-closed shape of the detection (leave on the first real PTS/DTS offset, refuse anything unproven), and the fixtures. The bounded rewind handling here is careful work and would have been the right shape had the axis been usable, so closing this in favour of #413 is about the remedy only, not the report or the analysis behind it.

superuser404notfound:
Superseded by #413, which repairs the same defect at the demuxer boundary and keeps the native path. The fixtures from this PR are the regression test there, and the reasoning is in the comment above.

PR #453, opened 2026-08-29
fix(audio): rebuild the resampler when a frame's input format changes mid-stream

### LINKED ISSUE #452: A live splice that changes the audio channel layout crashes the SW path: swr_convert reads a NULL input plane (the resampler is built once from the first frame)
**Environment:** AetherEngine 6.56.5 (`1463d707`), FFmpegBuild 2.5.0, tvOS 26.6 (23L773), Apple TV 4K 3rd gen (AppleTV14,1). Live TS source on the software path (`SoftwarePlaybackHost`, demux loop + `AVSampleBufferAudioRenderer`).

## What happened

~14 minutes into a live session the app crashed hard, right at a program boundary (21:20 local). The fault is on the demux thread, inside `swr_convert`:

```
Thread 7 CRASHED (engine demux queue)
  0 Libswresample   0x6490 (image offset)
  1 Libswresample   0x5484 (image offset)
  2 AetherEngine    closure #1 in AudioDecoder.appendFrameToPending(_:)
  3 AetherEngine    withUnsafeMutablePointer<A,B,C>(to:_:)
  4 AetherEngine    AudioDecoder.appendFrameToPending(_:)
  5 AetherEngine    AudioDecoder.decode(packet:)
  6 AetherEngine    demuxIteration #1 () in SoftwarePlaybackHost.runDemuxLoop(...)
```

`EXC_BAD_ACCESS` / SIGSEGV `SEGV_ACCERR`, `si_addr = 0` (Apple arm64 reports the protected zero page as ACCERR, so this is a NULL dereference, not a wild pointer).

## The faulting instruction

Both swresample frames sit below the first exported symbol (`swr_get_class` @ 0x7c10), i.e. in the stripped `audioconvert.c` area. Disassembling FFmpegBuild 2.5.0's `Libswresample` (arm64, UUID `CF55A27B-D7EE-3CD4-9789-28A3FDE89B39`, verified against the device image) at the fault pc:

```
0x6488  sxtw  x9, w2        ; src stride
0x648c  sxtw  x10, w3       ; dst stride
0x6490  ldr   s0, [x1]      ; <- fault, x1 == 0
0x6494  str   s0, [x0]
0x6498  add   x11, x1, x9
```

Registers at the fault: `x1 = 0x0` (source plane), `x2 = 4` (src stride: one packed float plane), `x3 = 0x18` (dst stride: **24 bytes = 6 ch × 4 B float interleaved**), `x25 = 0x2` (consistent with channel index 2 — the first plane a stereo frame no longer carries, though I can't prove that register's role from a stripped build).

So: the copy loop of `swri_audio_convert` is interleaving a **6-channel** output from **planar float** input and reads a **NULL input plane**.

## Root cause

`AudioDecoder` builds its `SwrContext` lazily from the first decoded frame and never revisits it:

```swift
while avcodec_receive_frame(ctx, f) >= 0 {
    if swrContext == nil {
        if !initResamplerFromFrame(f) { continue }
    }
    appendFrameToPending(f)
    ...
}
```

Nothing compares later frames' `ch_layout` / `sample_rate` / `format` against what the context was configured for. Live TS providers change the audio configuration mid-stream at program/ad splices (5.1 ↔ stereo is routine on European broadcast). FFmpeg's decoder follows the new config and emits frames with the new layout — but swr still reads input per its stale 6-plane configuration, and `frame->extended_data[2..5]` of a stereo frame are NULL → this crash, deterministically, at the first spliced frame.

Barring heap corruption (nothing else in the process was misbehaving), a NULL plane reaching that loop has no other route.

Related non-crashing damage from the same gap: a sample-rate or format switch under a built context resamples garbage instead of crashing — e.g. an HE-AAC stream whose first frame decodes at the core rate before SBR resolves stays pinned to the wrong rate for the whole session.

## Suggested fix shape

Per received frame, when `swrContext != nil`, compare the frame's `ch_layout` (via `av_channel_layout_compare`), `sample_rate`, and `format` against the configured input parameters. On change:

1. `emitPending()` first — the accumulator's bytes are in the OLD output format and must not be mixed or dropped;
2. `swr_free` + re-run the existing `initResamplerFromFrame(f)` (also rebuilds the `CMAudioFormatDescription`, so the renderer sees the new ASBD on subsequent buffers);
3. reset the gapless `AudioClockAnchor` — its emitted-sample count is denominated in the old rate; the next buffer should re-anchor from its container PTS.

Same guard in `drain()`.

I'm running exactly this fix on-device via a local override right now and will report back; happy to open a PR if you want it.

tschuegy:
Fixes #452.

`AudioDecoder` builds its `SwrContext` lazily from the first decoded frame and never revisits it. A live splice that swaps the stream's audio configuration mid-session (5.1 ↔ stereo at a program boundary is routine on European broadcast TS) leaves the context configured for the first frame's layout; a context built for more input planes than the frame carries reads a NULL plane inside `swr_convert` and crashes the demux thread. Full fault analysis (disassembly, registers, KSCrash capture) is in the issue.

**The fix:** record the input parameters the context was built for (`ch_layout` copy, `format`, rate) and compare each received frame against them in `decode()` and `drain()`. On a change:

1. `emitPending()` first — the accumulator's bytes are in the OLD output format and must be neither dropped nor mixed with the new format's;
2. `swr_free` + re-run the existing `initResamplerFromFrame(_:)`, which also rebuilds the `CMAudioFormatDescription` so the renderer sees the new ASBD on subsequent buffers;
3. reset the gapless `AudioClockAnchor` — its emitted-sample count is denominated in the old rate; the next buffer re-anchors from its own container PTS.

A failed rebuild leaves `swrContext` nil and skips the frame; the next frame retries, mirroring the lazy-init failure mode. The layout compare only runs when the frame carries a valid layout, so an UNSPEC stream that fell back to the synthesised default at init does not rebuild per frame. A rate/format switch that previously resampled garbage silently (an HE-AAC first frame at the core rate, say) now rebuilds too.

**Field-verified on device** (Apple TV 4K 3rd gen, tvOS 26.6, this branch as a local override on 6.56.5): tuned the crashing channel, hit a real on-air AC-3 5.1 → stereo splice two minutes in —

```
21:50:11.110 [AudioDecoder] Resampler ready: 48000Hz, 6ch, inFmt=8
21:51:51.735 [AudioDecoder] input format changed mid-stream (48000Hz/6ch/fmt=8 -> 48000Hz/2ch/fmt=8); rebuilding resampler
21:51:51.735 [AudioDecoder] Resampler ready: 48000Hz, 2ch, inFmt=8
```

One detection, no flapping; the SW diag line holds `dclk=1.00 / status=rendering` straight through the seam, playback continues cleanly. On 6.56.5 this exact moment is the crash from the issue.

https://claude.ai/code/session_01HewBbYmNEut2WvoEHmeBRU

### DIFF
--- a/Sources/AetherEngine/Audio/AudioDecoder.swift
+++ b/Sources/AetherEngine/Audio/AudioDecoder.swift
@@ -42,6 +42,14 @@ final class AudioDecoder: @unchecked Sendable {
     private var _loggedZeroConvert = false
     #endif
 
+    /// Input parameters the current `swrContext` was built for (AE#452). `decode()`/`drain()` compare
+    /// each frame against these and rebuild on a mid-stream change: a live splice can swap the
+    /// stream's audio configuration (5.1 <-> stereo at a program boundary), and a context built for
+    /// more input planes than the frame carries reads a NULL plane inside swr_convert.
+    private var configuredInLayout = AVChannelLayout()
+    private var configuredInFormat: Int32 = AV_SAMPLE_FMT_NONE.rawValue
+    private var configuredInRate: Int32 = 0
+
     private(set) var sampleRate: Int32 = 0
     private(set) var channels: Int32 = 0
 
@@ -99,6 +107,9 @@ final class AudioDecoder: @unchecked Sendable {
             // most and recovers immediately.
             if swrContext == nil {
                 if !initResamplerFromFrame(f) { continue }
+            } else if inputFormatChanged(f) {
+                if let sampleBuffer = rebuildResampler(for: f) { results.append(sampleBuffer) }
+                if swrContext == nil { continue }
             }
             appendFrameToPending(f)
             if pendingSampleCount >= Self.minSamplesPerBuffer {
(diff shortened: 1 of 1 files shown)


superuser404notfound:
Merged as `654c69e4`. Thanks for this one: the fault analysis was complete enough to review the fix against, and the field capture of the actual on-air splice is what a report like this usually never gets.

Reviewed and run against the full suite here before the merge (2374 tests, green), then hardened in `3c33907a` on top:

- The compare and the plane check moved into one type, `ResamplerInputParameters`, with unit tests around the cases the crash walked through: the reported 5.1 to stereo splice, a rate change, a format change, a layout change at the same channel count, and a frame that states no layout (which must not rebuild the synthesised default on every frame).
- The layout comparison now treats an undecidable result (a negative AVERROR, one of the layouts invalid) as changed rather than unchanged, matching what `AudioBridge` already does. `av_channel_layout_compare` never returns negative in the FFmpeg we ship, so this is contract, not behaviour, but continuing on a stale context is the exact fault being fixed.
- `AudioTapDecoder` had the same gap and is fed from the same live sources, so it rebuilds on the same signal now. Its output is pinned to mono 48 kHz, so nothing has to be flushed across the seam.
- Both decoders also refuse a frame that is missing a plane the context would read. Your compare covers a declared change; corrupt live MPEG-TS decodes to `nb_samples > 0` with a NULL plane and declares nothing, which is the same dereference by another route (`AudioBridge` already guarded that case on the loopback path).

Full suite green after the follow-up (2382 tests). A release carrying both goes out shortly.

Issue #96, opened 2026-07-03
Subtitle side-demuxer takes ~55 s to emit the first cue after a backward seek (video plays normally meanwhile)


rrgomes:
Split from the #93 retest.

Test asset: native Dolby Vision Profile 8.1, 4K HEVC Main10 (HDR10 fallback), ~25 Mbps, MKV, embedded text/PGS subtitle track (stream 2). tvOS 26.5, Apple TV 4K (3rd gen), loopback-HLS native path.

After an aggressive **backward** seek, the embedded subtitle reader blocks for ~55 s before emitting its first cue, while video plays normally at the new position the whole time — so subtitles simply disappear for ~a minute. Re-selecting the subtitle track in the UI doesn't help: it restarts the same reader, which blocks again.

Trace (backward seek to ~560 s at `t0`; video plays 560→575 fine throughout):

```
t0+3s   embedded subtitle reader started: stream=2 startAt=560.30s effectiveStart=562.50s seekTo=560.50s …
t0+56s  subtitle first cue: pktPTS=561812 → startTime=561.812s endTime=562.813s
t0+59s  embedded subtitle reader parked: demuxPos=666.0s playhead=576.0s lead=90s
```

So the reader takes ~55 s to surface the first cue at 561.8 s, then catches up in a burst (560→666 in the last ~3 s). The block correlates with the main video producer doing segment restarts / cache-misses on the same seek — suggesting the subtitle side-demuxer is starved on the shared source read (or reads forward from a stale anchor, cf. the closed #52).

During normal forward playback the reader scans cleanly (starts/exits roughly every ~1 s, emitting cues, parking at `lead=90s`) — the pathology is specific to re-seeking backward.

Happy to attach the full trace.

superuser404notfound:
Thanks for splitting this out, the trace shape is very telling.

[...]

Issue #27, opened 2026-06-08
Support IOReader in Probe API


strangeliu:
### What should the engine do?

AetherEngine now supports opening any custom IOReader, but the Probe API — `public nonisolated static func probe(url: URL, options: LoadOptions) throws -> SourceProbe` — still only accepts a URL. We should add a new overload (or extend the existing one) to accept MediaSource directly, so that custom I/O sources can be probed just like regular URLs.

### Motivating media or use case

Able to load SourceProbe info from any input io

### Area

Other

### Would you be willing to open a PR?

No

superuser404notfound:
Shipped in 3.0.0: `probe(source:)` accepts a `MediaSource`, so custom readers get the same one-shot metadata read as URLs. Ownership note: the caller keeps the reader; the probe seeks/reads through it, leaves the cursor at an unspecified position, and never calls `close()`. Hand `load(source:)` a fresh or rewound reader afterwards. Closing; feel free to reopen if the contract doesn't fit your use case.

PR #465, opened 2026-09-02
fix(demuxer): pace request rate after origin refusals

sitepilotusa:
## Measurement

On one TorBox link, eight parallel 32 MB reads and eight back-to-back 32 MB reads all served. The first 429 arrived after 36 sequential 256 KB reads in 10 seconds. An Apple TV seek then issued 11 requests in 4.4 seconds, including six 4 MB detour fills that bypassed the shared origin budget. Concurrency reduction alone did not lower request rate.

With the first two commits on an Apple TV, refusals fell from roughly 150 to 21 per session and seeks landed in 2.0 and 3.4 seconds. Every refusal after the pump went quiet came from the best-effort subtitle forward prefetcher reopening its side reader every 4 to 10 seconds. One play-press seek still stalled after 8 seconds and completed 9.6 seconds late through AVPlayer recovery.

With `6.60.0-atlas.2`, refusals fell again to 15 and seven of eight seeks landed in 0.04 to 7.5 seconds. The eighth seek issued three pump requests in 2.1 seconds while the previous refusal was 45 seconds old; the third request was refused and the seek stalled for 35 seconds. Refusals then continued every 2 to 4 seconds for 17 seconds until the reader rebuilt its demuxer.

The same `6.60.0-atlas.2` run showed two launches waiting 5.3 and 9.5 seconds for the origin chain's first byte—15 of one 22-second Play-to-picture interval. The reader reported only the whole-chain duration even though URLSession already collected one transaction metric per redirect hop, leaving the slow hop unknown.

## Change

- Arm a per-origin two-token bucket refilling at one request every two seconds after 429/503/509.
- Honor parsed Retry-After, otherwise use a 2/4/8/15-second quiet-period ladder.
- Disarm pacing only after 60 seconds without a refusal, without changing the learned concurrency limit.
- Put detour block reads through the same origin ticket path.
- Hold the speculative subtitle forward prefetcher while the shared origin is paced or serial; recheck without charging its restart budget.
- For tasks slower than one second, log each redirect hop in order with host and port only, optional HTTP status, rounded TTFB, and total time.
- Keep the API unchanged; no LoadOptions field is added.

The README note identifies the downstream public fork and can be omitted from the upstream merge.

## Verification

- `swift test`: 2,456 tests in 333 suites passed.
- Deterministic injected-clock tests cover pacing lifetime, redirect-chain lookup, quiet periods, Retry-After, timeout, the two-token/0.5-per-second refill and cap, the quiet-only disarm rule, and host concurrency precedence.
- The subtitle prefetch hold decision covers all four paced/serial combinations.
- Pure formatter tests cover the strict one-second threshold, hop order, host-only output, omitted status, and millisecond rounding.
- Private reader-loop integration remains the Apple TV device lane.

Pending device pass: Apple TV, alone on the link, traced: same scenario, plus two or three extra seeks a minute apart after the first refusal. Expected: no seek stalls past its first attempt; refusals stay in the single digits and none arrive in the 60 s after the previous one. On any slow load, grep for `[AVIOReader] slow first byte:` to identify the slow redirect hop without exposing its signed path or query.


sitepilotusa:
Closing; keeping this work on the fork.

superuser404notfound:
Thank you for this, and for the device traces that came with it. Measurements like "36 sequential 256 KB reads in 10 seconds before the first 429" and "11 requests in 4.4 seconds, six of them detour fills that bypassed the shared budget" are the part that cannot be reconstructed from here, and they are what made the request-rate case decidable at all.

The pull request is landing in pieces rather than as a whole, because its parts have quite different risk and one of them collides with work that went in while it was open. Here is the full accounting.

## Merged

**Scrub stills on VideoToolbox** — `34a14461`, via PR #470. Taken on its own since it is independent of the request-rate work. The guarding is what made it easy to take: hardware only for the internal `DataIOReader` the segment-cache still path uses, a latched fallback so a transfer failure cannot retry per frame, and issue #27's software default preserved for every public custom-reader and network caller.

**The request-rate pacer** — `7fa12915`, via PR #471, with one change of substance described below. This is the pacer, the detour block reads on the same origin ticket path, the subtitle forward prefetcher holding while the origin is paced or serial, and the redirect-hop first-byte log.

That one took two rounds. It merged on a green local run and failed CI in four tests that have nothing to do with pacing: two waiters reported "the waiter never got a thread; nothing about the budget was measured", a HEAD size probe never ran, and a served-from-memory count came back one high. It was reverted (`ef108d28`) while that was worked out.

The cause was the admission order. `acquire` took the concurrency slot first and waited for a rate token inside it, so on an origin capped at one request a paced caller parked every other path against that origin for the whole quiet period, up to fifteen seconds. Five existing tests drive real refusals and paid that ladder in wall time, 6 to 55 seconds each against under two before, and with 352 suites running in parallel enough threads sat in a blocked `acquire` that unrelated suites could not get one. The token is now taken before any slot: it is consumed rather than held, so a caller waiting on the rate rule occupies nothing. The wait is one timed wait in 250 ms slices rather than a 10 ms poll, what the pacer spends comes off the slot's budget so a caller's `timeout` is its total rather than that figure per gate, and the path where a slot wait timed out no longer skips the pacer, which it did before because the pacer only ran on the paths that got a slot.

**One of the assertions is reversed as a result.** "the request owns its concurrency slot while the pacer holds it" is the behaviour that broke CI, so it now claims the opposite, plus the ceiling-of-one case it exists for. Its one-second budget is raised to thirty so the injected clock decides when it is released rather than the wall clock on a loaded machine.

## Not taken

**The resident-ranges commit.** This landed separately as #468 while this pull request was open, with a follow-up: the spans are folded from the playlist axis onto the published display axis before they are published. Those two axes are unequal by construction on exactly this path, since AE#270 anchors the presentation origin on the container's start while the published shift also carries the producer's drift, so raw plan seconds sit beside the host's own scrubber. The version here predates that, and merging it would put the band back on the wrong axis and restore a synchronous main-actor read of the restart lock and the cache condition.

**The README note** is left out of the picks, as the pull request itself proposes.

## Held, with a question

**The delivery-rate range sizing.** The request cadence argument is sound and the measurement behind it is convincing. What is holding it is memory rather than cadence: it raises the resident window from 16 MB to 128 MB with a 64 MB low-water mark, and lets one range reach 2 GiB bounded only by a sanity limit. `window` is an in-memory buffer, and the primary target here is tvOS, where the per-app budget is tight enough that 287 MB resident is a jetsam question rather than a tuning one. A device measurement of peak footprint on an Apple TV, not a Mac, would settle it.

There is a second point worth a look independent of that: `PumpRangeSizeMemory` is process-lifetime and keyed by origin, so a size learned on a fast link is applied to the next session against that origin whatever the link is doing then.

**The far backward-seek re-anchor.** Independent of the request-rate work and in AE#69 / #93 / #96 territory, so it deserves its own round rather than riding along with this one.

Both are still on the `atlas` branch and nothing is lost. Happy to take either as its own pull request, or to carry the range sizing here once there is a device footprint number to put against it.

superuser404notfound:
Everything merged from this PR is released in [6.67.0](https://github.com/superuser404notfound/AetherEngine/releases/tag/6.67.0): the per-origin pacer, with rate admitted before concurrency (`d6a45e3c`), the redirect-hop first-byte line, and the hardware-decoded scrub stills. The range-sizing half stays on the `atlas` branch pending the Apple TV memory measurement, as discussed above.

PR #468, opened 2026-09-02
Publish resident segment ranges for host timelines

sitepilotusa:
## Why

A 2026-09-02 playback session ended with 64 resident segments spanning more than four minutes across several cache islands. AVPlayer loadedTimeRanges exposed only roughly twelve seconds around the playhead, then forgot a seeked-ahead island as soon as the playhead left it. This answers the upstream issue draft fourth ask for a buffered-ranges publisher with the cache-owned truth.

## Change

- Merge SegmentCache indexes into contiguous resident runs under its existing condition.
- Map VOD runs onto the same presentation axis as seek(to:); live publishes an empty array because DVR history has a separate contract.
- Publish AetherEngine.$residentRanges on the main actor, coalesced to at most four updates per second, and clear it on load and teardown.
- Document that residency is not a promise that a seek will be instant.

No other public API changes.

## Verification

- swift test on this upstream-main branch: 2,507 tests in 339 suites passed.
- Eight new tests cover empty, single, adjacent, separated and evicted cache runs; synthetic plan mapping; live-empty behavior; and store/evict publisher delivery.


superuser404notfound:
Merged. Thanks for this, and in particular for building it on the cache instead of on `loadedTimeRanges`: the measured session in the description is exactly the case where AVPlayer's answer is not the engine's, and the run-merge in `SegmentCache` is the right place for the truth to live. The 250 ms coalescer and the clear on load and teardown were both the correct calls, and the eight tests pin the shapes that matter.

Verified here before merging: `swift test` on the branch with current main merged in, 606 XCTest cases with 0 failures and 2551 swift-testing cases in 350 suites passed.

Follow-up pushed on top (`b15d034b`), with the reasoning, since three of the four points are things a reviewer can only see from inside this repo's history:

**The axis.** The plan is cut on the playlist axis, and that is not the axis the engine publishes. `currentTime`, `duration` and `seek(to:)` live on a 0-based display axis, and the fold between them is `+ playlistShiftSeconds - sourcePresentationOrigin` (`applyNativeHostClockTick`). Those two are not equal on the very path this feature serves: AE#270 deliberately anchors `sourcePresentationOrigin` on the container's start time while the published shift also carries the producer's drift, measured at -0.08 s on a B-frame MP4 whose first PTS is 0 and growing with each producer restart. So plan seconds published raw sit beside the host's own scrubber, by a little at first and by more after a few seeks. `onSeekStateChanged` already folds a segment time for the scrub target; the follow-up applies the same fold, seam-aware, and re-runs it when the producer publishes a new shift (a shift change moves the band without the cache moving at all). The engine-side method is renamed `residentPlaylistRanges()` so the axis is stated where the values are produced.

**Lock nesting.** `residentRanges()` held `restartLock` across `cache.residentIndexRanges()`, which takes the cache condition. Every other forwarder in that region snapshots the subsystem reference under `restartLock` and then calls out unlocked, and `contiguousForwardReadAheadSeconds` documents that property for itself. The plan is a COW array, so snapshotting it costs a retain and the nesting is not needed.

**Main-actor read.** `AetherEngine` is `@MainActor`, so the `didSet` on `nativeVideoSession` ran `session.residentRanges()` on the main actor, taking the restart lock and the cache condition right after `start()` while the producer is already storing into both. AE#422 is an open report of a main-actor engine read blocking for 13.3 s, so the follow-up asks for the first snapshot through `noteResidentSetChanged()` instead and lets it arrive off-main like every later one.

**Store hot path.** `Set(entries.keys)` twice per `store()` and per `adopt()` under the condition lock is O(N) on a path that runs once per segment, and `entries` reaches into the thousands when `forwardBufferSegments` is opened up (the clamp is 2700). Whether the index was newly inserted, and whether the prune doomed anything, are both already known at that point, so the same answer is available in O(1).

Four tests were added for the fold, including a guard case that the whole set fails against the un-folded behaviour. The doc entry moved into the Time table, next to `clock.$bufferedPosition` and `seekableLiveRange`, since that is where a host looks for an axis surface.

superuser404notfound:
Released in [6.67.0](https://github.com/superuser404notfound/AetherEngine/releases/tag/6.67.0), together with the two follow-ups: the band is published on the display axis the host scrubs (`b15d034b`) and its fold is timed off the cooperative pool (`8659cac1`). `$residentRanges` is documented in the Time table of `docs/api.md`.

Issue #158, opened 2026-07-20
VOD load option to reuse the native AVPlayer across loads so system PiP survives next-episode transitions


superuser404notfound:
## Context

Sodalite tvOS PiP (host-built AVPictureInPictureController around a second AVPlayerLayer, Sodalite feature/tvos-pip) works for start, restore, close, background keepalive (5.11.0), and subtitles. The one unsupported flow is next-episode auto-advance while the PiP window is open: the window closes mid-transition.

## Device evidence (Apple TV 4K 3rd gen, console-attached)

The next-episode reload builds a fresh NativeAVPlayerHost (log: #1 -> #2, fresh t+0.00 KVO stamps) with a new loopback session. Three host-side mitigations were tried and all failed on device:

1. Keeping the AVPictureInPictureController alive through the currentAVPlayer nil gap.
2. Keeping the old (stopped) player attached to the source layer through the gap and skipping AVKit re-binds.
3. Firing the advance 4s early so the swap lands while the old item still plays (no didPlayToEndTime before the swap; window still closed while host #2 was coming up).

The system closes the PiP window whenever its source layer's player is replaced or left without an active item. Platform behavior matches Apple's guidance: PiP survives item transitions only on the SAME player instance (AVQueuePlayer / replaceCurrentItem patterns).

## Requested capability

A load path (option or dedicated API) that keeps the current AVPlayer instance and swaps only the AVPlayerItem onto the new loopback session, ideally with the new item prepared before the swap (gapless handover, queue-style). NativeAVPlayerHost reuse infrastructure partially exists (stopInternal keepNativeHost, issue #15 MediaRemote re-registration), but a regular load() still ends in a fresh player as observed by the layer.

Host adoption (Sodalite): re-enable next-episode auto-advance during PiP once the player identity is stable across the transition. Until then the host disables auto-advance while PiP is active (v1, episode runs out inside the window).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NAPCg4tCDdSqkK6tPkNptw

PR #483, opened 2026-09-04
feat(native): let a host request the in-place item handover for a foreground replacement

Quick104:
## Summary

Lets a host that mounts the engine's own `AVPlayerLayer` request the AE#158 in-place item handover on a foreground episode change, instead of taking the nil-item gap that only PiP is protected from today. Also makes the `nativeRemoteHLS` bypass honour the handover, which it never did, even under PiP.

Refs #158. The host-side mitigations recorded there as failing are the ones we hit too.

## What changed

- `prepareForItemReplacement()`: a one-shot public request. The next `load()` consumes it, `stop()` cancels it, and a non-native outgoing session ignores it. `shouldHandOverItemInPlace` gains a `hostRequested` input; PiP behaviour is unchanged.
- When the item is kept across the load, `NativeAVPlayerHost.prepareForItemHandover()` retires the outgoing session's publishers before the engine subscribes for the successor, so the previous item's EOF, readiness, rate and clock are not replayed into the new session. The same-content #93 recovery swap keeps its clock as before.
- Main-actor hops queued by the outgoing item's KVO drop on the session guard rather than writing into the successor. `unloadCurrentItem` retires the session id up front for that reason.
- The `nativeRemoteHLS` bypass callsite now reads `pendingInPlaceItemHandover` like the loopback callsite does. Before this, a native->native remote-HLS load dropped the item to nil even while a PiP window was active.
- `CHANGELOG.md` entry, `docs/api.md` row, and four new cases in `PiPItemHandoverTests` (one-shot consume, non-native ignore, `stop()` cancel, PiP independence).

Why the host cannot solve this on its own side: the gap is between `stopInternal` detaching the item and the successor's `replaceCurrentItem`, and the engine owns both ends. We tried keeping the surface and layer stable through the gap and presenting the successor only after first frame; the layer still came back black on Apple TV. That matches the three host-side mitigations recorded as failing in #158.

## Test plan

- Device / OS: Apple TV simulator (tvOS 26) and iPhone 17 Pro simulator (iOS 26) through Silo; iPhone (iOS 26) on a signed build. Physical Apple TV 4K verification is scheduled and will be added to this PR.
- Source media: H.264 / AAC 5.1 MKV served as HLS remux and as HLS transcode (`nativeRemoteHLS`), 1080p SDR. Direct-play MP4 through the loopback path.
- Result: next-episode autoplay through a host-requested handover shows one `AVPlayerItem` swap and zero nil items on the shared player, same view and layer, successor first frame ready. Silo's regression test asserts exactly that on both routes; before the bypass change it recorded 2 item changes and 1 nil item on the remote-HLS route.
- `swift test`: 2634 tests, 360 suites; the one failure is `Issue458AudioLanguageMetadataTests` "three-letter labels ICU cannot name", which fails identically on unmodified `main` with Xcode 27 beta (ICU locale data). `PublicAPIDocumentationTests`, `DocumentedConstantsTests`, `PiPItemHandoverTests` pass. `Scripts/check-doc-links.py` clean. tvOS and iOS Simulator `xcodebuild` builds pass.

## Checklist

- [x] `CHANGELOG.md` updated

Built for [Silo](https://github.com/Silo-Server/silo-apple), a self-hosted media server client. Drafted with AI assistance (Claude Code); reviewed and tested by the author.


superuser404notfound:
Merged as 145d3c85. Thanks for this, and for writing up the reasoning rather than only the patch.

Verified here before merging: both `swift test` runners are green on the branch (606 XCTest with 0 failures, 2634 swift-testing in 360 suites), and the `Issue458AudioLanguageMetadataTests` case you flagged does not fail on this toolchain, which supports your reading that it is Xcode 27 beta ICU locale data rather than anything in the change.

The `nativeRemoteHLS` finding is worth calling out, because it was two defects rather than one. The bypass never passed `inPlaceSwap`, so the item went to nil as you describe. It also never consumed `pendingInPlaceItemHandover`, so the flag stayed armed and the next loopback `load()` would have spent it on an unrelated transition. Your consume-and-reset closes both.

One follow-up pushed on top (bb36f229): `layerReadyObservation` was the one observer left without the session guard the others now carry. That matters specifically on this path, since the handover leaves the outgoing item mounted on the layer through the gap, so it is the observer that keeps reporting there.

One thing to aim your Apple TV round at, from reading rather than measuring. `unloadCurrentItem(inPlaceSwap: true)` returns before the `avPlayer.pause()` that precedes the item swap. That pause is the issue #15 fix for audio leading video on episode autoplay: a reused player carries `rate = 1.0` across `replaceCurrentItem`, so the successor auto-resumes and beats the post-load `waitForSwitch` gate. On an iOS PiP window that costs nothing, because no panel switches. On tvOS with Match Frame Rate it is exactly the configuration the handover is for, and the gate is what keeps the first frame off a mid-transition panel. So the case to watch is a transition where the two episodes differ in frame rate or dynamic range: if the successor's audio starts while the panel is still dark, that is this, not the nil-item gap. If it reproduces, the fix belongs on the engine side and we will take it.

Issue #495, opened 2026-09-06
An origin behind a self-signed certificate cannot play, because no host trust decision reaches the engine's sessions or AVPlayer


RadicalMuffinMan:
### What should the engine do?

Give host apps a way to accept a certificate the system doesn't trust, so an origin behind a self-signed or private CA cert can play. Today the engine's owned sessions have no auth delegate and AVPlayer never asks anyone, so there's no host-side workaround at all.

### Motivating media or use case

Hi! I'm the dev for Moonfin, a third party Jellyfin client, and plenty of the users run their server behind a self-signed or private CA certificate. Right now there's no way for a host app to make one of those playable through the engine. I've been carrying a fork to do it and would rather upstream it than keep rebasing.

URLSession enforces system certificate trust. The in-demuxer network stacks the engine replaces never did, so a host whose own API layer bypasses trust ends up in a split state: browsing the library works fine and every engine fetch fails its handshake before a byte is read. The open surfaces as bare invalid data which reads like a corrupt file rather than a rejected certificate.

There's no host-side workaround for two separate reasons:

- The engine's owned sessions (disc reader, HLS ingest readers, audio tap fetcher) run without a delegate, so a host has nowhere to answer an authentication challenge.

- The native remote HLS route is worse. AVPlayer resolves a remote URL through its own networking. No delegate on the asset is ever asked about the certificate and an ATS exception doesn't cover it either so that route can't play such a server however the host configures trust.

I added two pieces to fix this, both off unless a host opts in.

- `EngineTLS.allowUntrustedCertificates` is read per challenge, so flipping it applies from the next connection without rebuilding sessions. A session level delegate covers the owned sessions that otherwise run without one, and doubles as the fallback for tasks that missed a per-task delegate. While the flag is off every challenge keeps the system's default handling, so nothing changes for anyone who doesn't opt in.

- Since no delegate reaches AVPlayer's own networking, we point the player at a loopback http address instead, which moves the https request onto a URLSession the engine owns and therefore back under `EngineTLS`. Playlists get rewritten on the way through so every variant, rendition, key and segment the player goes on to ask for arrives there too. Anything that isn't a playlist is relayed byte for byte. It's only constructed for sessions that opted in and is nil for everything else.

[...]

PR #506, opened 2026-09-06
feat(network): a host decides which origins it trusts

RadicalMuffinMan:
## Summary

Gives a host a way to accept a server certificate the system will not vouch for, so a media server behind a self-signed or private CA certificate can play. Answers per origin rather than per process, as discussed in the issue.

Refs #495 (the loopback stand-in follows in a second PR, so this doesnt close it)

## What changed

- **`EngineTLS.serverTrustEvaluator`**, `(@Sendable (URLProtectionSpace) -> Bool)?`, asked per challenge about the origin the challenge came from. nil, the default, keeps the system's default handling everywhere. Read per challenge rather than at session construction, so replacing it applies from the next connection without rebuilding sessions, and lock-guarded like `EngineLog.handler`.
- **A session-level delegate** for the owned sessions that ran without one, and a fallback for tasks that missed a per-task delegate. AVIOReader's four paths (probe, chunk, persistent, streaming) resolve through the same disposition.
- **Session coverage re-counted on current main**, not on the fork's base. Added `HLSCarriageProbe` and `RemoteHLSSubtitleProxy` to the six from the issue.
- **The live subtitle rendition fetch owns its session** instead of borrowing `URLSession.shared`, which cannot carry a delegate and was the one engine fetch no host decision could reach.
- **Docs**: a `Certificate trust` section in `docs/api.md`, and `Network/EngineTLS.swift` added to `PublicAPIDocumentationTests.hostFacingTypeFiles` so the naming rule covers this surface from here on.

Anything other than a server-trust challenge the evaluator accepted is left to default handling, so client certificates and HTTP auth are untouched.

## Test plan

- **Device / OS:** MacBook Pro (15,1, Intel i7-8850H), macOS 15.7.7, Swift 6.2.3
- **Source media (container / video codec + profile / audio / HDR-DV):** synthetic. A loopback HTTPS origin with a self-signed certificate for 127.0.0.1, run as a Python subprocess because NWListener rejects an imported in-memory identity with EINVAL. It serves Range requests from a constant 4 MB body and logs every request that crosses the handshake, which is what makes a refusal observable rather than inferred.
- **Result:** `swift test --filter EngineTLS`, 9 tests in 2 suites, all passing.
  - Resolver: no evaluator keeps default handling; a non-trust challenge never reaches the evaluator at all; the evaluator is asked about the origin the challenge came from; a declining evaluator keeps default handling; an accepted challenge with no evaluable `SecTrust` falls through rather than sending a nil credential.
  - Live handshake: no evaluator refuses and no request reaches the origin; an evaluator accepting this origin serves the reader 64 KB with the expected body byte; an evaluator that answers only for another host leaves this one refused.
  - The two refusal arms assert `AVIOReaderError.transportSecurityFailed` and not only that no byte crossed, so they are proof of the 6.69.0 classification as well.
- `swift build` clean, `swift test --filter PublicAPIDocumentationTests` green including the XCTest half.

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,28 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Added
+
+- **`EngineTLS.serverTrustEvaluator`: host opt-in to accept server
+  certificates that fail system trust evaluation.** Every engine fetch runs
+  over URLSession, which enforces system certificate trust that the
+  in-demuxer network stacks the engine replaces never did. A media server
+  fronted by a self-signed or private-CA certificate therefore keeps working
+  in a host whose own API layer bypasses trust, while the engine's open fails
+  its handshake before a byte is read. The evaluator is asked per challenge
+  about the origin it came from, so a host holding a LAN address behind a
+  private certificate and a WAN address with a real one answers for each, and
+  one that pins an SPKI hash decides for itself. nil by default; while nil,
+  and for every non-server-trust challenge, handling is unchanged. Covers
+  every session the engine owns: the AVIOReader probe, chunk, persistent and
+  streaming paths, the disc reader, both HLS ingest readers, the audio tap
+  fetcher, the carriage probe and the remote HLS subtitle proxy.
+
+### Changed
+
+- The live subtitle rendition fetch owns its session instead of borrowing
+  `URLSession.shared`, which cannot carry a delegate and was the one engine
+  fetch no host trust decision could reach.
 
 ## [6.70.0] - 2026-09-06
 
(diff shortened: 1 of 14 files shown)


superuser404notfound:
Merged. The per-origin closure is the right shape for the dual-URL case, and re-counting the session coverage against current main rather than the fork's base is what picked up the two that had appeared since, plus the `URLSession.shared` rendition fetch that no host decision could ever have reached.

The live-handshake suite is what makes this reviewable rather than plausible: a refusal that only says "no bytes arrived" proves much less than one that also names the classification, and asserting `transportSecurityFailed` on both refusal arms ties it to the 6.69.0 work at the same time. Thanks for that.

Issue #388, opened 2026-08-17
maxConcurrentSourceRequests binds the origin of the loaded URL only; after a 302 the pinned target is fetched under its own, uncapped key (#377 follow-up)


tschuegy:
Follow-up to #377, from reading 6.28.0 and main @3777214e rather than from a trace — I have not measured this against a real panel yet and can, if it helps narrow anything.

## Shape

An Xtream panel that advertises `max_connections=1`, loaded through the portal URL, which answers the media request with a 302 to a separate media host (a redirect that in this ecosystem is the rule rather than the exception, and the media host is what actually counts the connections). The host knows the provider allows one connection and says so:

```swift
LoadOptions(maxConcurrentSourceRequests: 1)
```

## What happens

The ceiling is registered once, for the origin key of the URL the host loaded (`AetherEngine.swift:2723`, `setHostLimit(_:for: url)`, keyed on scheme+host+port). Everything after the redirect runs under the media host's key, which has no host limit:

- At open, the tail prefetch and the optimistic `bytes=0-` pump both go out against the source URL (`open()`, `AVIOReader.swift:689/694`). The prefetch is correctly skipped (`requiresSerialRequests(source) == true`), the pump takes the **source** key's slot, follows the 302 and streams from the media host — with its ticket sitting on the source key for the connection's whole life (`PersistentReadDelegate.originURL` = the request URL it was opened with).
- The 200/206 pins the target (`persistentReceivedResponse(resolvedURL:)` → `recordResolvedURL`), so from here `requestURL()` is the media host. Every later ticket is taken for `request.url` = that pinned URL (`startPersistentConnection`, `syncRequest`, the size probe), i.e. under the target key.
- `detourEligible` asks `requiresSerialRequests(requestURL())` — the target key, uncapped — so a detour block is still opened as a second request while the pump is streaming from the same host. On a one-slot panel that is the 509, and only after it does `noteRefusal(for: target)` bring the target down to `max(1, peak/2)`; `peakInflight` there did not include the redirected pump, so the basis is whatever the detour alone reached.

Net: with the host declaring `1`, the media host still sees two concurrent requests the first time a backward/parse read needs a detour block (MKV cues, a non-faststart moov, a backward scrub), and the switch-off arrives only after that refusal — once per process, which is what the declared limit was meant to save.

[...]

Issue #316, opened 2026-08-07
nativeRemoteHLS: external subtitles can never become renditions (load-time dropped, post-load overlay-only)


edde746:
On the `nativeRemoteHLS` bypass, `LoadOptions.externalSubtitles` is dropped: `load()` returns through the bypass branch before the registration step, so a host that declares sidecars at load time gets nothing. The only way to get them into `subtitleTracks` there is `addExternalSubtitleTrack` after load, which `AetherEngine+Subtitles.swift:919-928` documents as **overlay-only — no native WebVTT rendition / PiP**.

That combination leaves remote-HLS sessions with no route to a rendition-backed external subtitle. It matters for Plex and Jellyfin transcodes, which are served as `.m3u8` and are told `subtitles=none`, so the stream itself carries no text: the sidecar is the only copy. Whenever the picture leaves the host's view hierarchy — PiP, AirPlay, a wired external display — the host overlay is suppressed and the subtitle simply disappears, while the same source on the probe-and-loopback path renders it.

Could `loadRemoteHLS` honour `LoadOptions.externalSubtitles` and declare them as legible renditions (`AVMutableMediaSelectionGroup` / the same declaration the loopback path makes)? Failing that, a documented way to ask for rendition eligibility on the bypass would let hosts choose between passthrough and external-route subtitles knowingly, rather than discovering it as a disappearing caption.

For context: choosing the loopback path instead is not a workaround here — the bypass is what preserves E-AC-3/Atmos passthrough for these sources.

superuser404notfound:
Both halves of this are confirmed, and the first one is worse than reported: the declaration was not just unused, it was dropped without a trace. `load()` routes a `nativeRemoteHLS` session into `loadRemoteHLS` and returns from there, well before the point where the probe path registers `LoadOptions.externalSubtitles`, so `subtitleTracks` came back empty with no error and no log line. Nothing distinguished a dropped option from a source that has no subtitles. The AE#154 reroute onto the same bypass dropped them too, and the legible-group discovery would have undone the registration anyway: it assigned `subtitleTracks` wholesale instead of merging, so on a source carrying its own renditions the sidecars would have been delisted a beat after appearing.

[...]

Issue #441, opened 2026-08-28
seekableLiveRange's lower bound is window arithmetic (max(0, edge − window)) — it never consults the cache, so the advertised rewind depth over-promises


cmcpherson274:
### What happened?

`seekableLiveRange`'s lower bound is `max(0, edgeTime − dvrWindowSeconds)` — pure window arithmetic that never consults the cache — so the advertised rewind depth over-promises whenever the retained content is shallower than the window, which on a fresh live session is always.

Measured in two regimes on a real live source (Jellyfin 10.11 `/stream.ts`,
`dvrWindowSeconds: 1800`, per-second sampling of `seekableLiveRange` /
`liveEdgeTime` / `behindLiveSeconds` over four sessions to 29 min):

- While `edgeTime < 1800`: `lowerBound` reads 0.00 for the entire run — including at
  session ages of 5–7 minutes when at most that much content can exist anywhere.
- Once `edgeTime > 1800`: `lowerBound` tracks `edgeTime − 1800` exactly, sliding 1 s/s.
- The REAL floor, probed by seeking: the session's own start plus ~19 s (a below-floor
  target lands at the floor — same clamp value across repeated probes in one session).
  The advertised bound and the real floor never met in any session we ran.

The code states the formula (`Video/LiveWindow.swift:17-20`, unchanged at cb3baaff):
`return Swift.max(0, edgeTime - w)...edgeTime`.

Consequences for a host building a rewind UI on this API:
- A scrub strip that derives its depth from `seekableLiveRange` over-promises massively
  (a 3-minute-old session with a 1800 s window advertises 30 minutes of rewind).
- `seek(to:)` targets inside the advertised-but-not-retained region are accepted and
  clamp silently (the seek's own completion reports the requested target; the truth is
  only visible on the next clock sample).
- `clampLiveResumeIfBehindWindow()` consumes the same bound for its DVR-window clamp
  (`lowerBound + 5`), so the engine's own resume clamping can aim at a position that was
  never retained.

Ask: bound `seekableLiveRange.lowerBound` by the cache's actual resident floor (the
segment cache knows its oldest resident output position), or expose the floor as its own
published value so hosts can intersect. Either shape makes an honest strip possible.

### Steps to reproduce

1. Load any live source with `isLive: true, dvrWindowSeconds: 1800` (native/loopback
   route).
2. Sample `seekableLiveRange` once per second. For the first 30 minutes of the session
[...]

PR #507, opened 2026-09-06
feat(network): the trust decision reaches AVPlayer through the remote HLS stand-in (#495)

RadicalMuffinMan:
## Summary

Second half of #495. The trust hook in #506 covers the sessions the engine opens, and AVPlayer's own networking is not one of them, so the native remote HLS route still could not play a self-signed origin. This mounts a relay on the stand-in #316 already puts in front of a remote master, so the engine makes the https request and the handshake happens where the evaluator is asked.

Stacked on #506. The first commit here is that PR; the second is this one.

Refs #495

## What changed

- **`HLSOriginRelay`**, mounted on `HLSLocalServer` rather than beside it. Allowlist seeded by the load and grown as playlists reveal where their sub-resources live, upstream fetch, playlist rewriting. No socket code of its own: the accept loop, session token, `0.0.0.0` bind and AirPlay LAN swap are the server's, so they come for free and are not reimplemented.
- **`HLSLocalServer`** takes an optional `relay`, its provider becomes optional so a relay-only session has none, and gains one route plus `sendRelay`, the only writer there that passes a status through from elsewhere and the only one that answers 206.
- **The two reasons compose.** `RemoteHLSSubtitleProxy.prepare(needsRelay:)` covers sidecars, relay, or both. With both, the #316 master carries the injected renditions and its origin variants come back through the relay. `rewritePlaylist(absoluteOnly:)` exists for exactly that: the injected renditions are named relatively and belong to the local server, so resolving them would send the player to the origin for a rendition only the engine has.
- **Relayed requests are charged to `OriginRequestBudget`.** Ticket around each fetch, `noteRedirect` when the final URL differs (#388), `noteRefusal` on 429/503/509 with `Retry-After`. Slot wait is 2.5 s, under AVPlayer's ~3.5 s first-byte window, and `acquire` is fail-open so a paced origin slows the session rather than failing it.
- **Ranges pass through untouched.** `Range` forwarded verbatim, never parsed or rebuilt, `Content-Range` and status mirrored. A rewritten playlist downgrades to 200 with no `Content-Range`, since its body no longer has the length the range described.
- The relay engages for any https origin once an evaluator exists, not only ones it would accept. Asking at load time means asking about a protection space no handshake produced, with no `serverTrust` for a host that reads one. The real question is put at the handshake the relay makes, so a declined origin fails there rather than being laundered. There's a test for that.

## On the four review questions

- **AirPlay.** Inherited from `HLSLocalServer`, which already binds every interface, carries the session token, and declares itself to `routeChangeNeedsReload`. Rewritten URIs follow the `Host` the request arrived on, so a receiver that reached the LAN address keeps fetching from it.
- **`RemoteHLSSubtitleProxy` overlap.** This is now one stand-in with two reasons to exist. The version I drafted before your review had a second socket server and silently dropped #316's renditions on a self-signed origin. Both are gone.
- **Rewrite ordering.** One rewriter, so there's no third order to compose. The #316 master is rewritten once, after `start()`, when the address exists.
- **Byte ranges.** Checked against the merged history rather than the issue titles. #468 is resident *segment index* runs for a scrub bar and shares no path with this. #465's range sizing was left on the contributor's branch per your merge note; what landed is the pacer, and the relay now participates in it. Before this fix a 429 seen through the relay never armed the pacer the reader and the subtitle prefetcher read.

## Test plan

- **Device / OS:** MacBook Pro (15,1, Intel i7-8850H), macOS 15.7.7, Swift 6.2.3
- **Source media (container / video codec + profile / audio / HDR-DV):** synthetic. Three loopback Python origins: a self-signed HLS origin serving a master, variant and 4 KB MPEG-TS segments; a self-signed ranged-body origin; and a plain-http origin that answers `Range` exactly as asked and logs what it saw. The last is what makes "forwarded verbatim" observable rather than asserted.
- **Result:** 22 tests across the three TLS and relay suites, plus 141 across the wider affected surface (`RemoteHLS*`, `HLSLocalServer*`, `AirPlay*`, `OriginRequestBudget`), all passing.
  - Through the relay: a client that never sees the certificate gets master, variant and a segment with the sync byte intact. No evaluator gets 502. An evaluator that answers for a different host also gets 502.
  - `bytes=100-199` reaches the origin as that exact string and comes back 206, 100 bytes, `Content-Range: bytes 100-199/4096`.
  - A 429 through the relay leaves `OriginRequestBudget.shared.isPaced(origin)` true.
  - An injected `subs_0.m3u8` survives `absoluteOnly` rewriting while the origin's variant is sent through the relay.
  - A request without this session's token is refused before it reaches the relay.
- `swift build --build-tests` clean. `PublicAPIDocumentationTests` green, both halves.

**Not tested:** AirPlay against a receiver. The iOS-only branch does not compile into the macOS test run, and I don't have a receiver here. The logic is the same as #316's, which is the strongest argument available, but it is inheritance, not a device.

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Public API changes are intentional and documented (none new; `EngineTLS` is documented in #506)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,28 @@ the public-API contract.
 
 ## [Unreleased]
 
-_Nothing yet._
+### Added
+
+- **`EngineTLS.serverTrustEvaluator`: host opt-in to accept server
+  certificates that fail system trust evaluation.** Every engine fetch runs
+  over URLSession, which enforces system certificate trust that the
+  in-demuxer network stacks the engine replaces never did. A media server
+  fronted by a self-signed or private-CA certificate therefore keeps working
+  in a host whose own API layer bypasses trust, while the engine's open fails
+  its handshake before a byte is read. The evaluator is asked per challenge
+  about the origin it came from, so a host holding a LAN address behind a
+  private certificate and a WAN address with a real one answers for each, and
+  one that pins an SPKI hash decides for itself. nil by default; while nil,
+  and for every non-server-trust challenge, handling is unchanged. Covers
+  every session the engine owns: the AVIOReader probe, chunk, persistent and
+  streaming paths, the disc reader, both HLS ingest readers, the audio tap
+  fetcher, the carriage probe and the remote HLS subtitle proxy.
+
+### Changed
+
+- The live subtitle rendition fetch owns its session instead of borrowing
+  `URLSession.shared`, which cannot carry a delegate and was the one engine
+  fetch no host trust decision could reach.
 
 ## [6.70.0] - 2026-09-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,28 @@ the public-API contract.
   streaming paths, the disc reader, both HLS ingest readers, the audio tap
   fetcher, the carriage probe and the remote HLS subtitle proxy.
 
+- **The remote HLS stand-in relays the origin so the trust evaluator reaches
+  AVPlayer.** `EngineTLS` only governs sessions the engine opens, and on that
+  route the origin URL goes to `AVURLAsset`, where AVPlayer resolves it through
+  its own networking and asks no delegate about the certificate. An origin
+  behind a self-signed certificate could direct play, since that reads through
+  AVIOReader, and then fail the moment it transcoded. Once a host has set an
+  evaluator, an `HLSOriginRelay` is mounted on the `HLSLocalServer` that #316
+  already stands in front of a remote master, so the https request is made by
+  the engine and the handshake happens where the evaluator is asked. Playlists
+  are rewritten so every variant, key, map and segment follows, and anything
+  else is relayed byte for byte with `Range` forwarded verbatim and
+  `Content-Range` mirrored. Relayed requests are charged to
+  `OriginRequestBudget` like every other fetch the engine makes, so a metered
+  origin arms the same pacer the reader and the subtitle prefetcher read.
+  Sessions with no evaluator set reach AVPlayer unchanged.
+
+- The trust opt-in and #316's subtitle renditions now **compose**. The
+  rewritten master carries the injected renditions and its variants come back
+  through the relay, so a self-signed origin with sidecars gets both. The
+  renditions the engine serves itself are named relatively and stay with the
+  server that owns them.
+
 ### Changed
 
 - The live subtitle rendition fetch owns its session instead of borrowing
--- a/Sources/AetherEngine/AetherEngine+LiveSubtitleRenditions.swift
+++ b/Sources/AetherEngine/AetherEngine+LiveSubtitleRenditions.swift
@@ -180,10 +180,16 @@ extension AetherEngine {
         return cues.filter { $0.endTime >= horizon }
     }
 
+    /// `URLSession.shared` cannot carry a delegate, so this was the one engine fetch a host's trust
+    /// decision could never reach. Owned and process-wide for the same reason the carriage probe's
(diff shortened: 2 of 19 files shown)


superuser404notfound:
Merged as it stood, thanks. All four review points are answered, and mounting on `HLSLocalServer` rather than beside it is a better answer than the one I had in mind: one stand-in with two reasons, and `absoluteOnly` drawing the line between the origin's variants and the renditions this engine serves itself.

Five follow-ups went in afterwards (3ae1483f, e6185677), all on the relay rather than on the design.

**Media is streamed rather than read whole** (a4beff22). Every relayed body was read to its last byte before anything was written. On a playlist that is necessary. On a segment it puts the whole download in front of the player's first byte, which is the one budget AVPlayer does not extend: the ~3.5 s (-12889) the pacer comment in this PR already cites. The quieter cost is worse on a ladder, because AVPlayer sizes the next rendition off what it measured, and behind a buffer it measures a loopback burst rather than the link. Bytes now go to the socket as they arrive, written by the server's worker thread, which was parked on the request anyway: the session's delegate queue is serial across every task on it, so writing from the callback would let one socket the player stopped reading hold up every other fetch in flight. Playlists, refusals and bodies of unstated length are still held. That last case also needed the playlist test asked again at the rewrite, since an unstated length fell into the playlist arm and the rewriter would have walked MPEG-TS as lines of text.

**The relay is mounted where the system refuses the origin** (bdf6afd2). Asking the evaluator at load time is not possible, agreed, and the reasoning for that is right. But the system can be asked, and its answer is the question: an origin the system trusts is one AVPlayer reaches unaided. One handshake with no delegate on the session, a one-byte range rather than a HEAD because origins that serve media answer the first and often not the second. Anything that is not a trust refusal answers "no relay", since an unreachable host fails the direct route too. Without this, a host answering for one LAN address moved every byte of every remote-HLS session through the process, including the origins with real certificates. The evaluator is unchanged and still decides at the handshake the relay makes.

**A refused certificate stays legible** (34633fd1). 6.69.0 reads the refusal off the failed item's `NSUnderlyingErrorKey` chain, and behind the relay the player's request went to loopback and came back a plain 502, so nothing in that chain says certificate any more. The relay remembers the handshake it lost and the item classification asks it.

**A blocking reload keeps blocking** (34633fd1). AVPlayer appends `_HLS_msn` / `_HLS_part` to a playlist URL that advertises `CAN-BLOCK-RELOAD` (#441), and those were read past along with the rest of the query, so the reload answered at once and the player asked again immediately. Every field other than `origin` rides onto the origin's own query now.

**A refusal is not rewritten into a service** (34633fd1). A 404 on a `.m3u8` was rewritten and framed as 200, which reaches AVPlayer as a parse error instead of as the one word it can act on. Only a success is rewritten. Same commit strips control characters out of a `Content-Type` or `Content-Range` before it is written into the response, since those are the two values an origin writes and a CR in one of them ends the header early.

One note on the test plan, because it is worth knowing rather than rediscovering. The 22 filtered tests are green here too, but `swift test` in full is a different reading twice over. `OriginRequestBudget.shared` is one object and suites run in parallel, so a `resetForTesting()` in another suite emptied the budget the pacer assertion was about to read; that other suite now takes an origin of its own instead of resetting globally. And the streaming proof had the same problem from the client side: timed through URLSession it passed locally in 0.14 s and read 20.5 s on CI with no body byte behind the head that did arrive, which is a shape the relay cannot produce. It reads off a socket now, which times when the relay put an answer on the wire and nothing else (150077a5).

PR #517, opened 2026-09-08
#407: extend the generated-PTS repair to AVI (measured XviD sample)

### LINKED ISSUE #516: AVI with packed B-frames: presentation timestamps derived from decode order (mpeg4/XviD)
## Summary

For AVI files carrying MPEG-4 ASP (XviD/DivX) with **packed B-frames**, the engine pairs decoded pictures with timestamps derived from decode order. Playback shows the picture stepping backwards several times a second while audio runs clean.

`aetherctl swdecode` diagnoses this itself, and its verdict already points at #407.

## Environment

- AetherEngine 6.73.0 (SPM)
- Observed on iPhone 13 Pro / iOS 26.5 (host app, `AetherPlayerSurface`)
- Reproduced headless with `aetherctl` on macOS 15.7, x86_64

The headless reproduction uses no host code, so this is not an integration artifact on my side.

## Reproduction

```
$ aetherctl swdecode --frames 120 file:///path/to/sample.avi

[SWDecoder] Opened: 720x304, codec=mpeg4, threads=8, 8-bit
[mpeg4 @ 0x…] Video uses a non-standard and wasteful way to store B-frames
('packed B-frames'). Consider using the mpeg4_unpack_bframes bitstream filter
without encoding but stream copy to fix it.

=== SW DECODER RESULT ===
Codec:                mpeg4 (id=12)
Source resolution:    720x304
Decoder open:         OK
Frames decoded:       38

=== FRAME TIMES (presentation order, first 38) ===
0.042  0.083  0.167  0.125  0.209  0.292  0.250  0.334  0.417  0.375  0.459  0.542
Steps backwards:      12 of 37
VERDICT: the pictures are paired with the wrong timestamps. The container withheld
         its PTS and something invented one from decode order (#407).
```

12 backwards steps out of 37 — roughly a third of frames. On screen this reads as the image jittering forward and back on any motion.

## Source characteristics

A 2000s-era XviD AVI rip. Nothing unusual beyond the packed bitstream:

```
codec_name=mpeg4
profile=Advanced Simple Profile
codec_tag_string=XVID
width=720  height=304
field_order=unknown
r_frame_rate=2500000/104271   (≈23.975)
```

Per-frame, over a 25-frame window mid-file:

- `interlaced_frame=0` and `repeat_pict=0` on every frame — the source is progressive
- `best_effort_timestamp_time` advances evenly at 0.0417 s with no gaps or duplicates
- frame types `I B P B P B B P …` — B-frames throughout

So the container's timing is not damaged and no telecine or interlacing is involved. The only unusual property is the packed bitstream.

## What I ruled out first

In case it saves someone time:

- **Interlacing / field order.** Toggling `deinterlaceMode` (`.auto` / `.software`) and `deinterlaceFieldRate` (`.field` / `.frame`) changed nothing, which is consistent with the deinterlacer never engaging — correctly so, since no frame carries `AV_FRAME_FLAG_INTERLACED`.
- **Decode path.** Forcing `preferredDecodePath = .software` changed nothing.
- **Frame rate / cadence.** Timestamps are evenly spaced; there are no duplicated or dropped frames in the source.

## Suggested fix

Two options, either of which should resolve it:

1. Apply the `mpeg4_unpack_bframes` bitstream filter when the AVI demuxer yields an MPEG-4 stream whose extradata/packets indicate a packed bitstream — this is what libavcodec's own warning recommends.
2. Take presentation time from the decoded frame (`best_effort_timestamp` / `AVFrame.pts`) rather than deriving it from packet order, for containers that supply no PTS.

For what it's worth, another FFmpeg-based iOS player handles this class of file by taking the timestamp from the decoded `AVFrame` rather than from the packet, with a source comment noting that AVI packet PTS order is unreliable for exactly these files.

## Why it matters

Packed B-frames were the norm for DivX/XviD AVI rips through roughly 2000–2008, so this affects a broad class of legacy files rather than one damaged sample. `docs/formats.md` documents AVI and MPEG-4 Part 2 as supported, which is what led me to expect these to play.

Happy to run further `aetherctl` commands against the sample or test a patch.

a1go3:
Fixes #516.

`VFWDecodeOrderPTSRepair` already describes this failure exactly; its gate was Matroska-only because no AVI had been measured:

> **Matroska only.** AVI carries a FourCC on every track and is DTS-only as well, but the `ms_compat`/`codec_tag` equivalence is established in `matroskadec` and nowhere else, and a VC-1 AVI has not been measured. A gate that is exact on one container beats a guess on two.

Here is the measurement. A 2000s XviD rip — `mpeg4` Advanced Simple Profile, tag `XVID`, 720x304, `video_delay = 1`, packed B-frames — takes the same transposition, and `aetherctl swdecode` reports it in the same words the Matroska case did:

```
before   0.042  0.083  0.167  0.125  0.209  0.292  0.250  0.334  0.417  0.375
         Steps backwards: 12 of 37
         VERDICT: the pictures are paired with the wrong timestamps. The container
                  withheld its PTS and something invented one from decode order (#407).

after    0.042  0.083  0.125  0.167  0.209  0.250  0.292  0.334  0.375  0.417
         Steps backwards: 0 of 37
```

## Why AVI needs no equivalence of its own

`avidec` has no second mode. The container carries no presentation timestamps at all, every track carries a FourCC, and the demuxer puts the frame index on `pkt->dts`. So on an AVI input any PTS present was necessarily invented by `+genpts` — which makes `codec_tag != 0` a tautology there rather than a signal, and leaves `video_delay > 0` carrying the whole gate.

The Matroska branch keeps its tag condition untouched: there the tag is what separates the VFW-carried tracks that withhold PTS from the natively mapped ones that must keep theirs.

Both existing exclusions hold unchanged in AVI, and are pinned by tests:

- zero reorder delay — decode order *is* presentation order, so the invented axis is the right one;
- H.264 / HEVC / AV1 — they can stay on the native path, where the fMP4 muxer refuses a timestamp-less packet outright.

## Ruling out the alternatives on this sample

So the diagnosis does not rest on the verdict alone:

- **Not interlacing.** `field_order=unknown`, and `interlaced_frame=0` / `repeat_pict=0` on every frame of a mid-file 25-frame window. Toggling `deinterlaceMode` and `deinterlaceFieldRate` on device changed nothing, consistent with the deinterlacer correctly never engaging.
- **Not cadence or a damaged rate.** `best_effort_timestamp_time` advances evenly at 0.0417 s with no gaps and no duplicates.
- **Not the decode path.** Forcing `preferredDecodePath = .software` changed nothing.
- **Not the packed bitstream itself.** It is the only unusual property of the file, but the pictures decode fine; libavcodec's `mpeg4_unpack_bframes` suggestion is left alone deliberately — a separate, independent improvement, out of scope for a gate change.

## Tests

`nonMatroskaContainer` pinned the old decision that `"avi"` must not arm the repair. It becomes `ptsCarryingContainer` over the containers that genuinely supply timestamps (`mov,mp4,…`, `asf`, `mpegts`), and four AVI cases join the suite: the measured XviD shape, tag-independence, the two shared exclusions, and per-element name matching for `isAVI`.

Full suite on this branch: **2743 tests in 373 suites, all passing.**

## Scope

The gate can now arm only on AVI, and only where `video_delay > 0` — files that today play with their pictures in the wrong order. Nothing that plays correctly changes behaviour.

Happy to adjust naming or split the test changes if you would rather keep `VFW` in the type name now that it covers a second carriage.

### DIFF
--- a/Sources/AetherEngine/Demuxer/Demuxer.swift
+++ b/Sources/AetherEngine/Demuxer/Demuxer.swift
@@ -717,7 +717,7 @@ public final class Demuxer: @unchecked Sendable {
         generatedPTSStreams.removeAll()
         guard let nameC = ctx.pointee.iformat?.pointee.name else { return }
         let formatName = String(cString: nameC)
-        guard VFWDecodeOrderPTSRepair.isMatroska(formatName) else { return }
+        guard VFWDecodeOrderPTSRepair.containerWithholdsPTS(formatName) else { return }
         for index in 0..<Int32(ctx.pointee.nb_streams) {
             guard let stream = ctx.pointee.streams[Int(index)],
                   let par = stream.pointee.codecpar,
@@ -731,7 +731,8 @@ public final class Demuxer: @unchecked Sendable {
             else { continue }
             generatedPTSStreams.insert(index)
             EngineLog.emit(
-                "[Demuxer] AE#407 stream=\(index) is VFW-carried (tag=\(fourCC(par.pointee.codec_tag)) "
+                "[Demuxer] AE#407 stream=\(index) in \(formatName) is FourCC-carried "
+                + "(tag=\(fourCC(par.pointee.codec_tag)) "
                 + "videoDelay=\(par.pointee.video_delay)); the container carries no PTS, so the "
                 + "+genpts axis is decode order. Clearing PTS, the decoder's reorder owns presentation.",
                 category: .demux
--- a/Sources/AetherEngine/Demuxer/VFWDecodeOrderPTSRepair.swift
+++ b/Sources/AetherEngine/Demuxer/VFWDecodeOrderPTSRepair.swift
@@ -37,11 +37,24 @@ import AetherLibavutil
 /// present on such a track was necessarily invented by `+genpts`. `video_delay > 0` narrows that to the
 /// streams where inventing it can transpose anything at all.
 ///
-/// Two deliberate exclusions:
+/// **AVI joins the gate (measured).** The exclusion above was for want of a sample; here is one.
+/// A 2000s XviD rip (`mpeg4`, Advanced Simple Profile, tag `XVID`, 720x304, `video_delay = 1`,
+/// packed B-frames) reaches `SoftwareVideoDecoder` with the same transposition, and `aetherctl
+/// swdecode` reports it in the same words:
 ///
-/// - **Matroska only.** AVI carries a FourCC on every track and is DTS-only as well, but the
-///   `ms_compat`/`codec_tag` equivalence is established in `matroskadec` and nowhere else, and a
-///   VC-1 AVI has not been measured. A gate that is exact on one container beats a guess on two.
+///     0.042  0.083  0.167  0.125  0.209  0.292  0.250  0.334  0.417  0.375
+///     Steps backwards: 12 of 37
+///
+/// AVI needs no `ms_compat` equivalence to establish the same fact, because `avidec` has no other
+/// mode: the container has no presentation timestamps at all, every track carries a FourCC, and the
+/// demuxer puts the frame index on `pkt->dts`. So on an AVI input any PTS present was necessarily
+/// invented by `+genpts`, which makes `codec_tag != 0` a tautology there rather than a signal, and
+/// `video_delay > 0` carries the whole gate. Progressive source, evenly spaced timestamps, no
+/// telecine: the packed bitstream is the only unusual property, and it is not the cause — the
+/// invented axis is.
+///
+/// One deliberate exclusion remains:
+///
 /// - **H.264 / HEVC / AV1 keep their PTS**, VFW-carried or not. Those are the codecs
 ///   `VideoRoutingPolicy` can route natively, which is the one path where packets reach the fMP4
 ///   muxer, and that muxer refuses a packet with no timestamp outright. A wrongly ordered picture is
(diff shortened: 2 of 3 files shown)


a1go3:
Verified on device: iPhone 13 Pro / iOS 26.5, host app built against this branch. The sample that produced #516 now plays with no visible judder — so the fix holds beyond the headless `swdecode` verdict, on the file the report started from.

superuser404notfound:
Merged, thank you. The diagnosis is right, the fix sits in the right layer, and the reasoning about `avidec` holds at the source: `libavformat/avidec.c` assigns `pkt->dts` and never once assigns `pkt->pts`, so `codec_tag != 0` really is a tautology on that input and `video_delay > 0` really does carry the whole gate.

Reproduced here against the FATE sample `mpeg4/packed_bframes.avi`, looped to 390 packets so the count means something:

```
main     Steps backwards:  64 of 239
#517     Steps backwards:   0 of 239
```

A full `aetherctl play` session on the same file plays through clean on the merged branch, `disp=+30 lost=0 dpts=33.4/33.4` at 29.97 fps, which is the check that mattered most: a picture that reaches the renderer without a timestamp is unschedulable there, and the decoder-side `best_effort_timestamp` repair does catch every one of them.

## One correction, pulled in afterwards as `8bffbb36`

The PR closes on "the packed bitstream is the only unusual property, and it is not the cause". That does not survive a second file. A plain coding-order XviD AVI of the identical shape (`mpeg4`, tag `XVID`, `video_delay = 1`, B-frames throughout) decodes in the right order on `main` today:

```
packed B-frames (FATE sample, looped)                64 backwards steps of 239
plain coding-order chunks (ffmpeg -c:v mpeg4 -bf 2)   0 backwards steps of 14
```

`+genpts` writes a coding-order axis in both. On a well-formed AVI that axis happens to be the presentation ladder shifted by exactly one frame, uniformly, so nothing changes places:

```
pkt   0      1      2      3      4      5        (dts = the AVI frame index)
dts   0.000  0.042  0.083  0.125  0.167  0.208
pts   0.042  0.167  0.083  0.125  0.292  0.208    <- each picture own slot, plus one
```

The packed bitstream is what breaks the correspondence that shift depends on. Its N-VOP placeholder chunks (7 bytes, no picture) hold a slot in the packet ladder while the picture they stand for rides inside the previous chunk, so packets and pictures stop matching one to one and the shift stops being uniform.

That changes nothing about the gate, which is why this went in as a follow-up rather than a review round. It matters for the next reader: someone comparing two XviD AVIs where only one is broken would otherwise have no way to tell them apart, and the widened gate would look like it was arming for no reason on the healthy one. Which it does, harmlessly, and that is now recorded too: with the axis cleared, `mpeg4` and `mpeg2video` in AVI with B-frames reproduce the same ladder frame for frame.

Also noted there is why the repair beats libavcodec own `mpeg4_unpack_bframes` suggestion, which was the right call to leave out of scope: clearing the axis covers every container that withholds PTS, where the bitstream filter would cover one codec.

## On the naming question

`VFWDecodeOrderPTSRepair` stays. AVI is the Video for Windows container and its stream header is a `BITMAPINFOHEADER`, so the name reads correctly for both carriages, and the log line saying "FourCC-carried" is the better place for the widening to show.

Suite is 2800 in 381 suites on `main`, both runners green.

PR #529, opened 2026-09-11
feat(dv): dolbyVisionHandling = .baseLayerOnly, the base layer of a Dolby Vision source

skrew:
Summary

Adds LoadOptions.dolbyVisionHandling (.automatic / .baseLayerOnly): a host option that presents the HDR10 / HLG base layer of a Dolby Vision source and leaves the Dolby Vision out of the container, on every display. The route a host offers as "Dolby Vision: off (HDR10)", for a title whose Dolby Vision is wrong and whose base layer is right.

The reported title is a 4K remux of 1917 whose container record claims Profile 5, compatibility 0 over a bitstream whose VUI declares BT.2020 YCbCr PQ (color_primaries=9 color_trc=16 color_space=9) and whose RPU carries disable_residual_flag=0, el_spatial_resampling_filter_flag=1 and NLQ (linear_dz), fields only a Profile 7 RPU has, with an identity mapping and HDR10 static metadata on the base layer. libavcodec warns Profile 5 RPUs should not use NLQ on it. Served as dvh1.05, as the record asks, AVPlayer reads YCbCr as IPT and the picture is green / violet on tvOS; every player that ignores the record shows the HDR10 base. Nothing in the container says which half is lying, so the choice is the host's, per title.

What changed

- DolbyVisionHandling and LoadOptions.dolbyVisionHandling (default .automatic, nothing moves for existing hosts). Declared next to forceDolbyVisionOnNonDVDisplay, listed in SessionOptionCorrection.knownFields at the same position (the Mirror order test), a tuning field: correctable through reloadAtCurrentPosition(applying:).
- VideoRoutingPolicy.dolbyVisionBaseLayerIsPresentable / presentsDolbyVisionBaseLayer: one predicate read by the codec route, the format clamp, the criteria request and the software-path guard. HEVC Profile 7 / 8.1 / 8.4 and AV1 10.1 / 10.4 are admitted by their record; a Profile 5 / AV1 10.0 record only when the VUI declares a BT.2020 YCbCr base with a PQ or HLG transfer (a genuine Profile 5 leaves both unspecified, IPT-PQ-c2 having no VUI code point). A Profile 5 whose VUI is unspecified keeps its route and the engine says so.
- CodecRoutePolicy: with the option, an admitted HEVC source is served as plain hvc1 (plainHEVCCodecs, the same string the P8.2 strip route uses), base-layer VIDEO-RANGE, no SUPPLEMENTAL-CODECS, doviConfig: .strip, no P7 conversion; AV1 takes the plain av01 branch with the record stripped. It runs ahead of every DV branch, so it wins over AE#455. On the default route a Profile 5 record over a YCbCr VUI is logged (DV Profile 5 record over a BT.2020 YCbCr VUI) so a host knows to offer the option.
- AetherEngine.load: the format clamp reads the session table with Dolby Vision left unclaimed (DisplayCapabilities.withoutDolbyVision()), so videoFormat resolves to .hdr10 / .hlg and the tvOS criteria ask for hvc1 rather than dvh1; sourceVideoFormat and sourceDVProfile keep naming what the file carries. The #176 software-path refusal and the VT-probe exemption stand down for a Profile 5 record the VUI contradicts, since that route plays plain HEVC.
- HLSVideoEngine: the P5 colour override keys on the sample entry the route chose (dvh1), not on the variant.
- aetherctl serve | validate | segverify | play --dv-base-layer, play --reload-applying dolby-vision=<automatic|baseLayerOnly>.
- DolbyVisionBaseLayerTests (16 tests: routes for 8.1 / 8.4 / 7 / relabelled 5 / genuine 5 / AV1 10.1 / 10.0, precedence over AE#455, the predicate table, the clamp, the software-path refusal, reload correctability). CHANGELOG.md, docs/api.md, docs/formats.md, docs/cli.md updated; the documentation tests pass.

Test plan

- Device / OS: Apple TV 4K, tvOS 26, Dolby Vision television; host app on the loopback route with engine-driven display criteria.
- Source media: MKV, HEVC Main10 3840x2160 23.976, dvcC Profile 5 / compat 0 / level 6, VUI BT.2020 / PQ / BT.2020-NCL, RPU disable_residual_flag=0 + NLQ (a Profile 7 remux relabelled 5), 11.8 GB over HTTP from an IPTV origin.
- Result, default route (unchanged): [HLSVideoEngine] DV source: profile=5 compat=0 … color_primaries=9 color_trc=16 color_space=9, the new DV Profile 5 record over a BT.2020 YCbCr VUI line, master served with CODECS="dvh1.05.06", picture green / violet on the panel, as before this PR.
- Result, .baseLayerOnly applied from the host's video settings (a reloadAtCurrentPosition(applying:) on the playing session, and on the following loads): presenting the PQ base layer only (dolbyVisionHandling=baseLayerOnly), plain hvc1, dvcC stripped, no SUPPLEMENTAL-CODECS, picture correct, host stats read source Dolby Vision P5 / panel HDR10.
- aetherctl validate on the first 8 MiB of the same file, macOS: default route logs the contradiction and serves dvh1.05.06; --dv-base-layer serves the plain hvc1 master.
- swift test: 2877 Swift Testing tests in 392 suites and 612 XCTest tests pass (1 skipped, the fixture-gated one).

Not measured: a genuine Profile 5 with the option (it keeps its route by construction, covered by the unit test only), AV1 Dolby Vision with the option (no hardware-AV1 box at hand), the HLG branch (8.4) on a panel.

Checklist

- [x] CHANGELOG.md updated
- [x] Commit messages follow Conventional Commits (feat(dv): …)
- [x] Public API changes are intentional and documented (docs/api.md › LoadOptions, docs/formats.md › Dolby Vision signaling, docs/cli.md)

A follow-up worth discussing separately: the contradiction is detectable without a false positive from the RPU itself (a Profile 5 RPU cannot carry a residual / NLQ), and the right automatic answer for that class is not the base layer but the existing P7 → 8.1 conversion, i.e. real Dolby Vision. That needs the first RPU parsed at probe time and a fixture, so it is left out of this PR.


superuser404notfound:
Merged as d45ce3f9. Thank you, this is a thorough piece of work: the predicate, the four call sites reading one answer, the log on the default route so a host knows to offer the option, and the docs all landed as written.

Verified here against Dolby's browser test kit before merging, since the mislabelled title itself is not reproducible on this end. Default route byte-identical on P5, P8.1 and P8.4. With the option, P8.1 serves `CODECS="hvc1.2.4.L150.b0"` with no `SUPPLEMENTAL-CODECS` and an `init.mp4` of 1322 B against 1354 B with no `dvvC` box in it, P8.4 keeps its HLG range, and the genuine Profile 5 (VUI unspecified, `color_trc=2 color_space=2`) keeps `dvh1.05.06`, which was the claim that mattered most. Full suite green on both runners after the merge with main (631 XCTest, 2883 swift-testing).

One follow-up pushed on top (9e4dba9e, comment only): the doc comment on `dolbyVisionBaseLayerIsPresentable` read as if 8.2 and 10.2 were left out, while the predicate table admits them, so the comment now says what the code does and why it costs nothing there.

The follow-up in your last paragraph is the right shape. A Profile 5 RPU that carries a residual or an NLQ is a contradiction the file states on its own, and converting that class to 8.1 is a better answer than dropping the Dolby Vision. Worth its own issue when the fixture exists.

superuser404notfound:
The follow-up from your last paragraph is built and on main: #532. A Profile 5 record over a BT.2020 YCbCr VUI now has its first RPU read, and a source whose RPU reads 7 or 8 is served as that profile, so the class gets real Dolby Vision through the routes that already existed rather than the base layer this PR falls back to. Your option keeps precedence, since a host asking for the base layer is answering a question the engine did not ask.

Measured on Dolby's Profile 8.1 signal relabelled to Profile 5 by two bytes. The Profile 7 arm is unit-tested only, because the reported title is not available here, so a run on your *1917* remux once this is released would be the one thing still missing. Details and the log line to look for are on #532.

superuser404notfound:
Released in [6.81.0](https://github.com/superuser404notfound/AetherEngine/releases/tag/6.81.0), along with the #532 follow-up. Both consumers are pinned to it. Thanks again for the option and for naming the better automatic answer in the same breath.

PR #533, opened 2026-09-13
fix(subtitles): preserve source axis for rebuilt VOD runs

orut34iop:
## Summary

Preserve the source-to-item timestamp normalization carried by rebuilt loopback-HLS VOD runs. Without it, raw-source-PTS subtitle formats such as PGS can be queried hundreds of seconds away from the visible playhead after a cached landing or recut.

## What changed

- Pass both the opening presentation shift and the persistent normalization shift from the segment producer to the engine.
- Track the opening and continuation source axes per rebuilt run, including backward epoch rewrites and zero-origin media.
- Use that source axis when reading cached landings and measuring rebuilt placements, while leaving existing-run placement behavior unchanged.
- Add standard Swift Testing coverage and an Unreleased changelog entry.

## Test plan

- Device / OS: Apple TV 4K (3rd generation), tvOS 26.6.
- Source media: MKV VOD, H.264 hardware path, embedded PGS subtitle track, two audio tracks, SDR; the media source timestamps begin about 600.2 seconds above the item timeline.
- Before: after a rebuilt/cached landing, playback at item 267–329 seconds queried subtitle coverage at item 252–313 seconds even though the corresponding PGS packets were on the source axis around 852–929 seconds, so both delivery and the overlay were empty.
- Result: on the same media, the selected PGS track remained visible continuously from 2:49 through 5:29 and after forward/backward seeks. The host application's physical-device acceptance passed.
- Focused: DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --filter RebuiltRunSourceAxisTests — 3 tests passed.
- Full: DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test — 2947 tests in 398 suites passed.

## Checklist

- [x] CHANGELOG.md updated
- [x] Commit messages follow Conventional Commits
- [x] No public API changes


superuser404notfound:
Merged as `5487b025`, thank you. The diagnosis holds and it is worth more than the title claims.

Reproduced here on macOS before touching the branch, on `Scripts/timecode-fixture.sh`'s
`tc-cues-lie.mkv` remuxed with `-output_ts_offset 600`, which gives the reporting geometry exactly:
item axis 0 to 120 s over a source that runs 600 to 720 s. Served over `Scripts/slowrange.py` at
600 kbps / 300 ms, the AE#481 arm (`--start-position 53 --seek-every 2 --seek-count 5
--seek-pattern 65,60,70,58,75`), with `--picture-probe` as the witness:

```
main   #481 ... opens at 72.000s, which is seg18's own playlist position, so it carries 0.000s
       capErr +0.017, cur=-515.10, VERDICT: clock did not advance
#533   #481 ... opens at 72.000s, which is seg18's own playlist position, so it carries 600.000s
       capErr -599.983 (one frame at 24 fps), cur=84.90, VERDICT: OK
```

So on that class of source it is not only the PGS query that misses: the published position goes
negative and the session reports a clock 600 s below the picture for the rest of the run. The
zero-origin control on the same arm is byte-identical between the two, which is the part that
matters for the nine rounds of AE#418 and AE#481 measurement that ran on fixtures starting at zero.
That is also what hid this: on those fixtures "this run has no placement offset left" and "item time
is source time here" are the same statement, and only one of them is true in general.

The gate-open pair states the whole thing in two lines, same bytes, same resume at 53 s:

```
zero origin   actual=43000  desired=52000 pinnedTo=43000 shift=0      presentedShift=-9000
600 s origin  actual=643000 desired=52000 pinnedTo=43000 shift=600000 presentedShift=591000
```

One gate, one backoff of 9 s, two numbers 600 s apart.

Two follow-ups on top, in `b3399179`, neither of them a correction to the reasoning here:

- The record was a second dictionary beside `epochShiftByIndex`, with the same key, call site, lock
  and pruning rule, and nothing to keep the two in agreement. Both quantities now live in one
  `EpochAxisTable` entry, with the pruning rule written once and the two lookups named for what they
  answer.
- The placement carries its own normalization instead of the reading looking it up again. A backward
  restart between a composition and its reading drops that epoch's entry, which is the pruning rule
  working as intended, and the late lookup would then have answered with an older run's number.

The AE#412 re-cut log line was reading from the same conflation and is fixed with it: it printed
"opened 591.000s below its boundary" for a gate that opened 9 s below it, because the shift it
quoted also carried the source origin.

superuser404notfound:
Released in [6.85.0](https://github.com/superuser404notfound/AetherEngine/releases/tag/6.85.0) (`21ab499`), and both consumers are on it.

One more thing came out of the same lens after the merge, so it is in that release too. The composition had the same conflation the landing rule had: it added the whole shift per placement rather than what the gate backed off, so on the 600 s twin a placement worth 596.833 s composed onto a standing 599.625 s and published **1196.458 s**, with the seam at item -531.625 s. It self-corrects wherever the placement can be read back out of AVPlayer's buffer, which is why the arm here recovered within a tick; AE#418 round 7 is the case where it cannot, and there it would stand for the session. The composition now works in AVPlayer's own displacement and the normalization is added once, by the epoch that wrote the bytes:

```
before   #418 seg17 placed ... axis shift 599.625s -> 1196.458s from item -531.625s
after    #418 seg13 placed ... axis shift 599.625s -> 590.625s from item 81.000s
         (the reading then lands on 591.000s, which is what the picture reads)
```

That also collapsed the own-run / rebuilt branch this PR introduced: once the composition is in displacement terms, a placement's bytes are worth the same on either timeline, and what the two disagree about is the base, which is measured.

The offset-origin fixture recipe is worth having if more of this comes up: `-output_ts_offset 600` on `tc-cues-lie.mkv`, offset applied BEFORE the Cues are injected (a `-c copy` remux into MKV regenerates them). Note `capErr` inverts on such a fixture, so -600.000 is the healthy reading there and 0.000 is the defect.

Issue #536, opened 2026-09-15
An outgoing AetherPlayerSurface rebinds the engine while it is removed, leaving the next session's layer attached to no view (black video over audio)


tschuegy:
### What happened?

A host that remounts its player view by identity while keeping the same `AetherEngine` instance ends up with an engine bound to no view, so the next session's layer is attached nowhere: the new item plays its audio over a black picture, and pause/play does not bring the picture back.

The host is Syravo (iOS/tvOS). Its "Next episode" button swaps the request on one engine, and the SwiftUI subtree holding `AetherPlayerSurface` is keyed with `.id(request.id)`, so the surface is torn down and a new one is made. Traced on an iPhone with a log line in `makeUIView` / `updateUIView` / `dismantleUIView` of a representable that binds exactly like `AetherPlayerSurface` (bind in make and in update), SwiftUI runs the swap in this order:

1. `makeUIView` for the **incoming** view → `bind(view: new)` detaches the old view, `boundView = new`
2. `updateUIView` for the incoming view
3. `updateUIView` for the **outgoing** view (it still observes the same host model while it is being removed) → `bind(view: old)` detaches the new view, `boundView = old`
4. `dismantleUIView` for the outgoing view → the async `detach()`; the view is released and the weak `boundView` becomes `nil`

The incoming view is never updated again (its `engine` input did not change), so nothing binds it. When the next `load()` builds the new `NativeAVPlayerHost`, `presentCurrentLayer()` runs against `boundView == nil` and returns. The new `AVPlayerLayer` reports `isReadyForDisplay == true` with `superlayer == nil` and a zero frame.

This is the sibling of #188: there the view was reused and the engine swapped; here the engine stays and the view is swapped, and the outgoing view's rebind (the #188 fix in `updateUIView`) is what steals the binding.

### Steps to reproduce

1. Keep one `AetherEngine`; show `AetherPlayerSurface(engine:)` inside a subtree keyed by `.id(token)`, where the subtree also observes an `ObservableObject` that publishes during the swap (any playback model does).
2. Play a native-path item.
3. `engine.stop()`, change `token`, `load()` the next item.
4. Audio plays, the picture stays black.

Engine-level equivalent, no SwiftUI needed:

```swift
let old = AetherPlayerView(frame: bounds)
engine.bind(view: old)
var new: AetherPlayerView? = AetherPlayerView(frame: bounds)
engine.bind(view: new!)   // incoming makeUIView
[...]

Issue #188, opened 2026-07-21
AetherPlayerSurface never rebinds on engine replacement — reused SwiftUI views leave a fresh engine with no bound view (black video over audio)


rrgomes:
`AetherPlayerSurface` binds the engine to its platform view only in `makeUIView`; `updateUIView` is empty. When a host app replaces its `AetherEngine` instance (a retry/reload flow that tears down and recreates the engine) at the same structural position, SwiftUI reuses the platform view, so the new engine's `bind(view:)` never runs. Its `boundView` stays nil, the post-load `presentCurrentLayer()` no-ops on the guard, and the reused view keeps displaying the previous engine's detached layer — black video while the new engine's AVPlayer plays audio. The `isReadyForDisplay` diagnostic reads true throughout, since the host's layer reports ready without being attached to any on-screen view.

Suggested fix: have `updateUIView` call `engine.bind(view: uiView)` (bind is already idempotent for `existing === view`, so steady-state updates are cheap), or document that hosts must key the surface's identity to the engine instance. We're shipping the `.id(ObjectIdentifier(engine))` workaround host-side; happy to retest a fix.

superuser404notfound:
Fixed in **5.18.4** (just released). Your diagnosis was exactly right.

`AetherPlayerSurface` only ran `bind(view:)` from `makeUIView`/`makeNSView`. On an engine swap at the same structural position SwiftUI keeps the platform view and calls only the update path, which was empty, so the new engine was never handed the surface: `boundView` stayed nil, the post-load `presentCurrentLayer()` bailed on its guard, and the recycled view kept showing the previous engine's detached layer while the new AVPlayer played audio. `isReadyForDisplay` reading true throughout lines up too, since the host layer reports ready without being on-screen.

The fix is the first option you suggested: `updateUIView`/`updateNSView` now call `engine.bind(view:)` on every update. `bind` is idempotent in steady state (the `existing === view` case skips the detach, `presentCurrentLayer()` re-attaches the same layer, and `attach` short-circuits when the hosted layer is unchanged), so per-update passes are a cheap no-op swap; on a swap the new engine takes over the reused view and re-attaches its layer. You can drop the `.id(ObjectIdentifier(engine))` workaround once you move up.

[...]

PR #537, opened 2026-09-15
fix(view): keep the picture when a surface is remounted on the same engine (#536)

tschuegy:
## Summary

The engine held one weak reference to its bound view, so a host that remounts `AetherPlayerSurface` by identity while keeping the engine could end up bound to nothing: the next session's layer attached nowhere and played audio over a black picture. The engine now keeps every bound surface weakly and presents on the most recently bound one that is still alive, falling back to the next one when that view is unbound or released.

Reported in #536

## What changed

- `AetherEngine.bind(view:)` / `unbind(view:)`: a weak list of bound surfaces in bind order replaces the single `boundView`. `boundView` is now the most recently bound live surface. Unbinding the presenting view detaches it and re-presents the layer on the fallback. Unbinding a non-presenting view only removes it, so it can never come back as a fallback.
- `AetherPlayerView.bindingEngine` (internal, weak): the engine that last bound the view. A second engine binding the same view takes it over from the first, which drops it from its surfaces and falls back to its own remaining one. This keeps #188's engine swap on a reused view working, and stops the first engine from reclaiming a view the second one holds.
- `AetherPlayerView.attach(_:)` / `detach()`: a previously hosted layer is removed only while it still sits in the view. Without this, the takeover above moved engine A's layer to its remaining surface, and engine B's attach on the taken-over view then pulled that layer back out of it, because the view still recorded it as its own (the takeover test caught this; the takeover also detaches the view first).
- `AetherPlayerSurface.dismantleUIView` / `dismantleNSView`: unbind synchronously through `bindingEngine` instead of `detach()` in a `Task`, so the layer moves to the incoming surface the moment the outgoing one leaves rather than at the next load.
- `Issue536SurfaceFallbackTests`: replays the measured SwiftUI order (bind incoming, rebind outgoing, dismantle outgoing) with and without an explicit unbind, the non-presenting unbind, and the takeover by a second engine.
- `docs/api.md` (`bind(view:)`, `unbind(view:)`) and `CHANGELOG.md`.

No public API changed.

## Test plan

- Device / OS: iPhone 16e, iOS 26.7, Syravo host.
- Source media: MKV, H.264 High 1920x1080 + AAC stereo, native loopback route. The defect is not format-dependent.
- Result: before, tapping Next Episode (same engine, surface keyed by `.id(request.id)`) played audio over black; a host-side probe read the new `AVPlayerLayer` as `isReadyForDisplay=true`, `superlayer=nil`, frame zero at +0 ms, +1.5 s and +5 s. The field fix was verified with Syravo's host-side equivalent of this change (a representable that rebinds after the swap): the picture appears, and the incoming view holds the layer. This PR's engine build has not been run on the device yet.
- On macOS: the four new `Issue536SurfaceFallbackTests` pass, and so do `Issue188SurfaceRebindTests` (the engine swap on a reused view is unchanged), `Issue120RemoteHLSLayerAttachTests`, `PublicAPIDocumentationTests` and `DocumentedConstantsTests`.
- Full `swift test` on macOS 27.0: 631 XCTest cases pass (1 skipped); 2975 Swift Testing tests run with one failure, `Issue458AudioLanguageMetadataTests` ("three-letter labels ICU cannot name stay unresolved", `iso639_2T(forSourceLanguage:)` returns "com"). It fails the same way on unmodified `main` at `56953aff` on this machine, so it is unrelated to this change and looks like an ICU difference on macOS 27.
- `xcodebuild -scheme AetherEngine -destination 'generic/platform=iOS Simulator' build` (compiles the UIKit surface): succeeded

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Public API changes are intentional and documented (none)


superuser404notfound:
Merged in edfde39c, thank you. The traced SwiftUI call order made this straightforward to review, and the tests replay exactly that order, including the takeover by a second engine that keeps #188 intact. Full suite is green on top of 6.89.0, and it goes out with the next release.

I changed `Closes #536` in the description to `Reported in #536`, so the issue stays open until this engine build (rather than the host-side representable) has run in Syravo on the device.

Issue #538, opened 2026-09-15
Renderer audio session activates on the main thread (hang-risk warning on iOS/tvOS 27)


tschuegy:
### What happened?

`activateRendererAudioSession()` calls `AVAudioSession.setActive(true)` synchronously on the main actor. On iOS 27 and tvOS 27 the system flags this at runtime:

```
AetherEngine+Loading.swift:1615 AVAudioSession_iOS.mm:978 This method can lead to UI unresponsiveness if called on the main thread. Consider using the asynchronous activate/deactivate API instead for calls from the main thread.
```

`setActive` is an XPC round trip to mediaserverd, the same reason the category declaration moved off-main in #114 and the teardown deactivation in #215. The renderer-path activation is the one call left on the main actor. I expected it to follow the same pattern.

### Steps to reproduce

1. Run a host on iOS 27 or tvOS 27 from Xcode, with the runtime-issues pane open.
2. `engine.load(...)` a source that takes a renderer path: the software host (`loadSoftware`) or audio-only (`loadAudio` / `loadAudioNative`). These are the three callers of `activateRendererAudioSession`.
3. The warning above points at `AetherEngine+Loading.swift:1615`.

### AetherEngine version or commit SHA

6.88.0 (`56953aff`)

### Host app

Custom / my own integration (Syravo)

### Platform

iOS and tvOS

### OS version

iOS 27.0 / tvOS 27.0

### Device / chip

iPhone and Apple TV 4K

### Playback path

Software (dav1d) host / audio-only

### Anything else

I have a PR ready that follows the #114/#215 pattern. The activation, the channel preference that depends on it, and the log line run in a detached `userInitiated` task, and the renderer load awaits it. The session is therefore still active before the host is built. The asynchronous `activate(options:)` API is iOS/tvOS 27 only, so it would need SDK gating in a package that supports iOS 16 / tvOS 17. Moving the synchronous call off the main actor is what the warning actually asks for.

superuser404notfound:
#539 is on `main` as b50c42d3, plus 5e4cca98, which queues the activation and the #215 teardown release so a `stop()` during a renderer load's activation cannot release the session before that activation. Both will be in the next release. Leaving this open until the runtime-issues pane stays quiet for a software-path and an audio-only load on iOS 27 and tvOS 27.

superuser404notfound:
[...]

Issue #114, opened 2026-07-07
AVAudioSession hang-risk diagnostic: setCategory runs on the main thread in AetherEngine.init()


dlev02:
# AVAudioSession hang-risk diagnostic from `AetherEngine.init()` (setCategory on main thread)

## What happens

Constructing `AetherEngine` on the main actor trips Xcode's runtime hang-risk diagnostic:

```
AVAudioSession Hang Risk
AetherEngine.swift:1096  SessionCore.mm:631
This method can lead to UI unresponsiveness if called on the main thread while the audio session is active.
```

Line 1096 is the category setup in `public init()`:

```swift
try session.setCategory(.playback, mode: .moviePlayback, policy: .longFormAudio)
try session.setSupportsMultichannelContent(true)
```

## When it fires

Any time a new engine is created while the shared `AVAudioSession` is already active — e.g. the second playback in an app session (the previous engine/AVPlayer activated the session), or when a route like AirPods is up. Hosts construct the engine on the main thread (it's effectively MainActor-bound), so the `setCategory` XPC round-trip to the audio server happens on the main thread. Under contention that call can block long enough to be a watchdog/hang risk — which is exactly what the diagnostic is warning about.

## Suggested fix

Keep the issue-#24 semantics (declare category early, never activate at init — AVKit/`activateRendererAudioSession()` drive activation), but move the `setCategory`/`setSupportsMultichannelContent` pair off the main thread — e.g. a small detached/utility task or the same queue that later runs `activateRendererAudioSession()`. Nothing reads the category synchronously at init time, so the only requirement is "category is set before first activation", which the renderer/AVKit activation path can await.

## Repro

1. iOS host app, engine pinned at `main` HEAD `7ea6d3e`.
2. Play any stream (session becomes active), dismiss the player.
3. Start a second playback (new `AetherEngine()` on the main thread).
4. Xcode 26 runtime-issues pane logs the AVAudioSession hang-risk diagnostic pointing at `AetherEngine.swift:1096`.

superuser404notfound:
Fixed on `main` (`e59d42b`), thanks for the precise diagnosis.

[...]

PR #539, opened 2026-09-15
fix(audio): activate the renderer audio session off the main actor

tschuegy:
## Summary

The software and audio-only paths activate `AVAudioSession` themselves in `activateRendererAudioSession()`. That was the last session call still running synchronously on the main actor, and iOS/tvOS 27 flag it as a hang risk. This moves it off-main the same way #114 (category) and #215 (teardown release) already did.

Reported in #538

## What changed

- `activateRendererAudioSession` is now `async`. It still resolves the source channel count on the main actor, because that reads published track state. The blocking part (`setActive(true)`, `setPreferredOutputNumberOfChannels`, the log line) moved into `nonisolated static applyRendererAudioSession(sourceChannels:)`, which runs in a detached `userInitiated` task that captures no engine state.
- `loadSoftware`, `loadAudio` and `loadAudioNative` await it before building their host. Activation therefore still comes before the host, as before. Each then calls `checkLoadCurrent(generation)`, so a load superseded during the wait unwinds without building anything.
- I did not use the asynchronous `activate(options:)`: it is iOS/tvOS 27 only and would need SDK gating in a package that supports iOS 16 / tvOS 17. The warning is about the main thread, and the synchronous call off it answers that.
- `CHANGELOG.md` has an `[Unreleased]` entry, and the renderer-activation sentence in `docs/formats.md` says where it runs.

## Test plan

- `swift build` (macOS): passes.
- `xcodebuild -scheme AetherEngine`, `generic/platform=tvOS Simulator` and `generic/platform=iOS Simulator`: both build.
- `Scripts/check-doc-links.py`: all links resolve. `swift test --filter 'PublicAPIDocumentationTests|DocumentedConstantsTests'`: 15 tests pass.
- Device / OS: the warning was observed from Syravo on an iPhone (iOS 27.0) and an Apple TV 4K (tvOS 27.0), pointing at `AetherEngine+Loading.swift:1615`. The runtime check with this branch on those devices is still to be done.
- Source media: any source that routes to the software host or the audio-only path (the three callers).

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Public API changes are intentional and documented (no public API change)


superuser404notfound:
Merged in b50c42d3, thank you. The PR follows the #114 and #215 shape closely, and the generation check after the wait covers the case of a load being replaced while it waits. It goes out with the next release.

One follow-up on my side: with the activation off the main actor, it and the #215 teardown release became two detached tasks with no order between them. While the activation ran synchronously on the main actor, a `stop()` could not land inside it, but now a stop during a software or audio-only load's activation could release the session first and leave it active after a final teardown. Both now go through one queue, with a test for each ordering: 5e4cca98.

As on #537, I changed `Closes #538` in the description to `Reported in #538`, so the issue stays open until the runtime check on the iPhone and the Apple TV has been done.

Issue #532, opened 2026-09-11
A Dolby Vision Profile 5 record its own RPU contradicts should be routed as what the RPU says


superuser404notfound:
Follow-up to #529, raised there by @skrew.

#529 gave a host the option to present the HDR10 base layer of a source whose Dolby Vision record is wrong (`LoadOptions.dolbyVisionHandling = .baseLayerOnly`). That is the right answer when nothing can tell which half of the container is lying. For one class it is not the best answer available, because the file states the contradiction itself.

## The class

A container record claiming Profile 5, compatibility 0, over a bitstream whose VUI declares BT.2020 YCbCr with a PQ or HLG transfer. A genuine Profile 5 leaves `transfer_characteristics` and `matrix_coeffs` unspecified, IPT-PQ-c2 having no VUI code point, so the pairing is already a contradiction and #529 logs it. The reported instance is a 4K remux of *1917* whose RPU carries `disable_residual_flag = 0`, an enhancement-layer resampling filter and NLQ, fields only a Profile 7 RPU has.

## Why the RPU settles it

A Profile 5 RPU cannot carry a residual or an NLQ, so reading one is not a heuristic, it is a proof that the record is wrong. libdovi already answers this in one field, `guessed_profile` on `dovi_rpu_get_header`, and the engine already calls that function for `DoviRpuConverter.enhancementLayerType`.

Measured here against real media:

| Source | Container record | RPU says |
|---|---|---|
| Dolby test kit Profile 5 | profile 5 compat 0 | `dovi_profile: 5`, `vdr_rpu_profile=0`, `disable_residual=true` |
| A real Profile 8.1 remux | profile 8 compat 1 | `dovi_profile: 8`, `vdr_rpu_profile=1` |
| The reported title | profile 5 compat 0 | 7 (residual + NLQ) |

## What should happen

When the record says Profile 5 and the VUI contradicts it, read the first RPU. If it says 7 or 8, believe the RPU and take the route that profile already has: the Profile 7 to 8.1 conversion on a display with Dolby Vision, the HDR10 base on one without, and a truthful `dvcC` on the way out through the existing compatibility rewrite. Real Dolby Vision, rather than the base layer #529 falls back to.

The audit is gated on the contradiction, so a genuine Profile 5 and every other source read no packets at all. `dolbyVisionHandling = .baseLayerOnly` keeps precedence, since a host that asks for the base layer is answering a question the engine did not ask.

[...]

PR #567, opened 2026-09-20
feat(dv): recognise a recordless Dolby Vision Profile 5 from its first RPU

oxyc:
## Summary

Some Matroska releases are Dolby Vision Profile 5 but carry no DV record in the container (and an
unspecified VUI), so the engine loads them as plain SDR and the IPT picture is decoded as YCbCr: a
violet/green cast. This reads the first RPU for that narrow case and adds the missing record, so the
existing Profile 5 paths take over.

## What changed

- For a probed HEVC stream that is 10-bit 4:2:0, has no `AV_PKT_DATA_DOVI_CONF`, and has transfer, matrix
  and primaries all unspecified, the demuxer reads the first RPU (`rpuProfileOfSource`, libdovi's
  `guessed_profile`). Only if it reads exactly profile 5 does it add a Profile 5 record (level 6, RPU
  present, no EL, BL present, compat 0) to the stream's `coded_side_data`. Route, display criteria and
  the segment muxer's `dvcC`/`dvh1` then follow the existing Profile 5 paths; the load, the HLS producer's
  own open and rebuilds agree with no verdict to carry.
- Narrow on purpose (the #532 principle): any named colour description, 8-bit, non-HEVC, or an existing
  record never reaches a packet read; a failed or non-5 read leaves the source as it was. URL sources only
  (non-live, non-disc); cost is one extra open and up to 16 packets, for candidates only.
- A recordless P5 routed to the software path now meets the existing #176 refusal, like a real P5 record,
  instead of playing with a cast.
- Separate commit: skips the P5 `colr` override for a synthesized record (the source deliberately says
  nothing about colour; a plain `ffmpeg -c copy -tag:v dvh1` copy carries no `colr`). It is not shown to be
  necessary, only that playback is correct without it; happy to drop it.

## Test plan

- Device / OS: Apple TV 4K (3rd generation, AppleTV14,1), tvOS 27.0 (24J361); also `aetherctl` on macOS.
- Source media (container / video codec + profile / audio / HDR-DV): Matroska / HEVC Main 10, 3840x2160,
  full range, colour tags unspecified / AC-3 5.1 / Dolby Vision Profile 5 with no container record
  (streaming-service WEB-DL; every frame carries a profile 5 RPU).
  - Synthetic repro (test pattern, 3 s, HEVC with profile 5 RPUs and no container record):
    recordless-p5-synthetic.mp4. `aetherctl probe` reports `format: sdr` on unmodified `main` and
    `AE#recordless: DV profile 5 ...` / `format: dolbyVision` with this change. It shows the detection, not
    the colour cast.
  - The real file is copyrighted so it isn't attached; happy to email a real clip where the green tint is
    visible, on request.
- Result: before, `prepared: codec=hvc1 ... range=SDR DV=none`. After, `AE#recordless: DV profile 5 from
  the first RPU`, `DV source: profile=5 compat=0 level=6`, `prepared: codec=dvh1.05.06 ... range=PQ
  DV=profile5`, master `CODECS="dvh1.05.06" VIDEO-RANGE=PQ`, and the served `init.mp4` carries `dvh1` >
  `hvcC`, `dvcC`. Played on the Apple TV with correct colours. 7 new unit tests; 32 pass, 4 skipped for
  fixtures not in-repo.
  
https://github.com/user-attachments/assets/1152ec9f-bc4d-4a21-851e-9cf5df369ecc

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Public API changes are intentional and documented (none)


superuser404notfound:
Merged, thank you. The detection is the right shape: the record goes onto the demuxed stream before classification, so the route, the display criteria, the segment muxer's `dvcC` and the #176 software refusal all follow with no verdict to carry, and because the hook sits behind the `skipStreamInfo` guard in `probeStreams` the subtitle and Atmos side demuxers and the audit's own second open never reach it, so it cannot recurse and those passes pay nothing.

Reproduced here before merging, on a recordless Profile 5 built from Dolby's own P5 asset (the `dvcC` block addition mapping renamed so libavformat drops the record, bitstream untouched): `main` reports `format: sdr`, the branch reports `AE#recordless: DV profile 5 from the first RPU` and serves `CODECS="dvh1.05.06"`, `VIDEO-RANGE=PQ`, `init.mp4` with `dvh1` > `hvcC`, `dvcC`. Full suite green.

Two things I pulled in afterwards rather than sending the PR back.

**The `colr` commit is reverted.** That override is the AE#19 repair, and it exists for exactly the signal shape this detection gates on: a Profile 5 whose VUI says nothing gets the canonical IPT-PQ-c2 tuple written as `colr nclx` so AVPlayer's Dolby Vision decoder engages on the `dvh1` sample entry. Measured on one bitstream in two containers, the same 8 s of Dolby's P5 asset with and without its record: with the record, `init.mp4` is 883 B and carries `colr nclx` 9 / 16 / 9 with the full-range flag; with the record stripped, your branch serves 864 B and no `colr`. That makes the same bytes two different deliveries depending only on what the container happened to write down. If the green cast you saw with the override is reproducible on your file, it is worth its own issue rather than a branch of the new path: every recorded Profile 5 gets that identical stamp today, including a full-range one, so it would be a defect in the shipped route. A pair of runs on your file, current main against main with the `colr` skip put back, would settle it, and the skip goes back in if it shows (27648d2).

**The recordless arm reads fewer packets.** 16 is right for #532, where the record-over-VUI pairing already proves the source is broken before a packet is read. Here the budget is spent mostly by sources that were never Dolby Vision: an untagged 10-bit HEVC SDR encode is a common shape. Measured against an origin with 150 ms of latency and 1 MB/s, on a 30 MB untagged 10-bit SDR MKV, `probe` goes from 0.50 s on main to 2.75 s on the branch, to 1.21 s with a budget of 4 and to about 1.2 s with a budget of 2. A Profile 5 answers in the first video packet by construction, and the recordless Profile 5 fixture still does at 4, so that arm now reads 4 (44c9b66) and #532 keeps its 16.

Thanks for the write-up as well: the synthetic repro and the before and after lines made this reviewable without the source file.

Issue #107, opened 2026-07-06
AU live tv


tresby:
### What should the engine do?

Would love to be able to use this with OTA HDHR Australian streams.

They require deinterlacing - would be cool to use the hardware yadif_videotoolbox video filter for this.

And remix/transcoding of DVB_Teletext subs. These subs can be quickly remuxed/transcoded to text or ass format with libzvbi using -text_format text (or ass) and -fix_sub_duration (ffmpeg cli). Would mean instead of big ugly bitmap subs - these could be displayed as native subs via WEBVTT.

Let me know what you think and if there is any way I can help.

### Motivating media or use case

Australian OTA/FTA HDHR TV compatiblity.

### Area

Codec / container support

### Would you be willing to open a PR?

Yes

superuser404notfound:
Thanks for the detailed request, and for offering to help.

I scoped this against the engine and there is good news on the deinterlacing side. AetherEngine already runs a software bwdif/yadif deinterlacer on its software decode path, and MPEG-2/VC-1 already route there, so those deinterlace today. The gap is interlaced H.264 (typical for 1080i broadcast), which currently stays on the native AVPlayer path and is not deinterlaced. I am adding a rule that routes interlaced H.264 to the same software deinterlacer, so no hardware yadif_videotoolbox filter is required. Software bwdif is motion adaptive and actually higher quality, and it covers every interlaced codec with one path.

For the subtitles: DVB teletext needs libzvbi, which is not currently compiled into our FFmpeg build. I am rebuilding FFmpeg with libzvbi and wiring the teletext decoder to emit text (txt_format=text, txt_page=subtitle), so captions render as native text subtitles instead of bitmaps.

To validate against your exact streams, could you share:

1. ffprobe output for one AU channel, to confirm the video codec and field order:

   `ffprobe -hide_banner -show_streams -select_streams v:0 <stream>`

2. Whether the captions are DVB teletext (and which page, often 801 in AU) or DVB subtitles (bitmap). The subtitle stream line from:

   `ffprobe -hide_banner -show_streams -select_streams s <stream>`

   would settle it.

A short sample capture (even 15 to 30 seconds with a subtitle visible) would let me verify teletext timing and rendering end to end. No pressure on the PR offer; the stream details alone are a big help.

superuser404notfound:
Quick update: both halves are now implemented and merged.

- Interlaced H.264 now routes to the software bwdif deinterlacer, so 1080i broadcast plays without combing.
[...]

Issue #416, opened 2026-08-24
After a re-aimed restart, a PGS set whose clear is stranded below the producer floor overstays at the landing


rrgomes:
On the 6.40.0 re-aim path (same 1080p H.264 ~35 Mbit/s MKV + PGS SDH asset as #408), a seek that re-anchors and restarts the producer can land showing the previous position's caption over the new scene. Reproduced on 6.40.0; 6.41.0's diff does not touch the subtitle store, so it should hold there. Two devices, same day, same build:

- Apple TV 4K 3rd gen: a coalesced far seek to 205 landed on target; the overlay then applied a set from half a minute earlier:

  ```
  [applySubtitleEvent] cueStart=33.200s cueEnd=4295000.495s sourceTime=15.256s engine.currentTime=205.000s seekGen=2
  ```

  The next authored set on the stream is 215.674, so the stale one stands ~10 s over the new scene's audio.

- M4 Pro Mac (controlled repro): drain seek target 199.1; `boundary 196029 is not a random-access point (no sync sample covers it)`, re-aimed ×2, gate opened at `actual=191983`, 12.30 s flight. Post-landing, the picture at ~200 s renders under the 187.729 set (authored ≈2 s, a sound-effect caption):

  ```
  [applySubtitleEvent] cueStart=187.729s cueEnd=4295155.024s sourceTime=199.400s engine.currentTime=199.400s seekGen=10
  ```

  No stored subtitle packet exists in (191.98, 215.674). It cleared by ~206 on this hardware; the overstay scales with device speed and with the gap to the next authored set (we have watched the same shape run tens of seconds on the Apple TVs).

Mechanism as we read it: the re-aim's floor (191.98) lands past the 187.73 set, so that set's clear (~190) sits below the floor and can never be re-harvested — the store prunes behind the producer. The retained cue keeps `end_display_time = UINT32_MAX`, or the next-stored-packet close reaches the far-side successor (215.674) — a real packet, but not this set's successor, which is the #362 round-2 shape. `closeOpenEndedCues(startingBefore:)` does not cover it: the set starts inside the reconstruction window, not before it. Fast landings never show it — two control runs stepping +10 s with sub-second landings kept every caption correct.

Two candidate shapes, your call: close a bitmap set at `min(next stored packet, start + a PGS dwell cap)` when its own clear is unreachable; or a one-shot subtitle-stream harvest of [floor − gap, floor] after a re-aim, so the authored clear is read before the region is abandoned.

[...]

PR #573, opened 2026-09-20
fix(playback): carry a seek target onto the source axis (#107)

tschuegy:
A seek on a source whose first packet is not at zero rewinds to the start of the file, and then the reader starves with nothing reported.

## What happens

`SoftwarePlaybackHost.seek(to:)` takes the **session axis** — the axis the host publishes `currentTime` on — and hands it unchanged to four things that speak the **source's own timestamps**:

- `Demuxer.seekBounded(to:)`
- the packet store's `beginSeek` / `prepareSeek` / `endSeek(sourceClock:)`
- `videoDecoder.skipUntilPTS` and `renderer.setSkipThreshold`
- `AudioOutput.seekClock(to:)`

For a zero-based source the two axes coincide, so nothing was ever wrong. For a mid-stream-joined one — the case #107 introduced `sessionZeroSeconds` for — they differ by however far into the broadcast the source was opened, and two things break at once:

1. The demuxer is asked for a timestamp that lies *before* the file's first packet and clamps to the beginning, so the seek rewinds.
2. `reservoirSecondsLocked()` measures `storedVideoSeconds - sourceClock`. With the store on the source axis and `sourceClock` on the session axis, the reservoir reads as the entire offset. The producer concludes it has hours of buffer, relaxes, and stops reading; the consumer then starves. Nothing fails, so nothing is logged and no recovery path can fire — the picture simply stands while the clock runs on past the end of the file.

## The change

One conversion at the top of `seek(to:)`, applied to everything that speaks the source axis. `currentTime` stays on the session axis.

`SWClockAnchorPolicy.sourceSeconds(forSession:sessionZeroSeconds:)` is the inverse of the mapping the host already applies when publishing the position, and the identity when the offset is zero. `liveScrubStill` in the same file already makes exactly this conversion with `sessionStartPts`, and says so in its doc comment — the VOD path was simply missing it.

## Test plan

- `swift test` — 3147 tests green, including four new cases on the conversion: identity without an offset, the carry itself, a round trip against the position the host publishes, and passing a target through rather than making it worse when it cannot be expressed.
- **Device:** Apple TV 4K (3rd generation), tvOS 26. A 64.5 s MPEG-TS capture on an SMB share whose first PTS is 24549.835 s, played through `AetherEngineSMB`.

| | before | after |
|---|---|---|
| packet cache | `result=miss resident_bytes=0` | `result=hit resident_bytes=44565599` |
| audio lead | `aLead=24549.8` | `aLead=1.96 … -0.04` |
| frames after the seek | `enq=+0 disp=+0` on every tick | plays on from the new position |

Seeking forwards and backwards repeatedly, and resuming at a stored position, both behave.

## How it was reached

Syravo now records live television to an SMB share and plays it back through `AetherEngineSMB`, which is what produced a source with a non-zero timestamp origin in the first place. Happy to adjust the shape of the fix if you would rather the conversion sat somewhere else.


superuser404notfound:
Merged, thank you. The diagnosis is exactly right, and the write-up made it possible to check rather than take on trust.

Three things were worth verifying before this went in, and all three hold:

- `Demuxer.seekBounded(to:)` wants an absolute source timestamp. It hands `seconds * AV_TIME_BASE` straight to `avformat_seek_file`, and the time-seekable path subtracts `sourcePTSOriginSeconds` itself, so the carry does not double-apply.
- `SoftwarePacketReadAhead` really is on the source axis on both sides: `updatePlayhead` is fed the raw synchronizer clock, and `prepareSeek` matches against keyframes stored in source PTS. That is the cache miss and the reservoir reading as the whole offset, as described.
- `clockSessionZero` is the right variable to carry with, and this is the subtle part: it is `firstSample - initialSeconds`, not `firstSample`. So the mapping stays correct after a resume as well, where the session axis has its zero point before the file's first packet.

Two changes were made on merge, neither of them to the logic: the CHANGELOG entry was rebased onto the two that landed ahead of it, and three em-dashes were replaced, since the house style does not use them.

One thing the review turned up that this PR does not cover, mentioned here because it is the same root cause. `AetherEngine.swift` states the subtitle pump's harvest anchor at the session-axis target while `subtitleDrainTick` reports that reader's progress from `sourceTime`. On a mid-stream-joined source the anchor lands the whole offset below every note that follows it, and `noteReach` has no unannounced-advance guard, so one run stretches over ground nobody read and the #416 landing gate stops refusing anything. That is #575.

This will go out in the next release; there is no need to run anything further on it in the meantime.

Issue #560, opened 2026-09-19
Capability request: record a live stream to a file from the session's existing connection


UnknownCoder807:
### What should the engine do?

Provide a public API to record the currently playing live stream to a file chosen by the host, fed from the connection the session already has open, so the host does not need to open a second connection to the origin. Something along the lines of startRecording(to: URL) / stopRecording(), plus a published recording state (idle / recording / failed with a reason such as disk full or source reset).

### Motivating media or use case

IPTV live channels (MPEG-TS over HTTP and live HLS), loaded with LoadOptions(isLive: true), with and without dvrWindowSeconds.

Users expect a Record button on live TV. Today a host can only do this by opening its own second connection to the stream and muxing it separately. On IPTV that is a real problem: most provider plans cap an account at 1 to 3 simultaneous connections, so a second connection for recording either fails outright or knocks the viewer off the channel they are watching. The engine already holds the one permitted connection and already demuxes every packet, so it is the only place a recording can be taken without paying for a second one.

The existing DVR timeshift does not cover this. It is a sliding window that evicts (SegmentCache under aether-segments/, sized by LiveWindowSizing), so it lets the viewer rewind but nothing survives the session, and there is no public way to keep or export what is in it.

What would make it useful in practice:
- Stream copy of the source packets (no re-encode), taken before any audio bridging, so the file keeps the original video and audio codecs. Output as MPEG-TS or fragmented MP4; either is fine, as long as a file that was cut short by a crash or a kill is still playable.
- Recording continues while the viewer pauses or scrubs inside the DVR window, because it follows the source rather than the playhead.
- A defined behaviour when liveSourceReset fires or the host reloads the same URL: either the recording carries on into the same file, or it ends cleanly and reports that it ended, so the host can start a new part.
- Errors reported through the API rather than silently (disk full, write failure, source ended).
- Nice to have, not required: "record from the start of what is buffered", i.e. include what is already in the DVR window when recording starts, so a viewer who presses Record a minute late does not lose the beginning.

### Area

Public API surface

### Would you be willing to open a PR?

Maybe, with guidance

superuser404notfound:
Implemented in #570, and thank you for the report: the connection-limit argument is what made this an engine problem rather than a host one, and it is quoted almost verbatim in the API docs.

**What it does**

```swift
try await engine.startRecording(to: url)
await engine.stopRecording()
engine.$recordingState        // .idle / .recording / .ended / .failed
```

[...]

PR #574, opened 2026-09-20
fix(recording): start a recording at zero, not at the broadcast clock (#560)

tschuegy:
A recording written by `LiveRecordingWriter` carries the broadcast's own clock, so its first presentation timestamp lies hours past its own beginning.

## What happens

The writer copies the source's `pts`/`dts` verbatim. A live channel's clock is wherever it has got to — on a stream that came up in the morning, six or seven hours in — so the resulting file starts at, say, 24549.835 s rather than at zero.

That is legal, and this engine copes with it on the VOD path since #107. But a recording on a NAS is opened by whatever the household happens to use, and a player that treats the first timestamp as the origin reads the file as hours of nothing before the picture. A duration probe reports the offset rather than the length.

It is also what turned #107's missing axis conversion into a hard freeze — see the companion PR — which is how this surfaced at all.

## The change

The origin is latched once, from the first packet written (the arming keyframe), preferring `dts` because that is the order the muxer writes in. The same instant is subtracted from **every** stream, so the A/V relationship is preserved exactly.

- `AV_NOPTS_VALUE` travels through untouched, as `OutputTimestampSanitizer` treats it.
- A reordered frame sitting just before the arming keyframe's DTS clamps to zero rather than going negative, which the muxer would refuse. At most the first few frames share instant zero; nothing is reordered.
- Nothing is decoded and nothing is re-encoded. Only the numbering changes, so the "never decodes, never encodes" property of the writer still holds.

## A question rather than an assumption

If the broadcast timeline is worth keeping, this is the wrong trade and I would rather you said so. A metadata tag would carry the original origin without putting it on the presentation axis, and the bytes would stay byte-identical to the source. Say the word and I will redo it that way — I could not tell from the code whether the verbatim copy was a decision or an oversight.

## Test plan

- `swift test` — 3146 tests green, including three new cases: the shift itself, `AV_NOPTS_VALUE` and a packet with no origin yet passing through untouched, and the clamp for a reordered frame. The existing container assertion still holds, so the output is still a whole number of 188 byte packets.
- **Device:** Apple TV 4K (3rd generation), tvOS 26, recording a live MPEG-TS channel whose clock stood at 24549.835 s to an SMB share.


superuser404notfound:
Merged, thank you, and the question at the end is the right one to ask: no, the broadcast timeline is not worth keeping on the presentation axis. A recording is a file, and a file begins at zero. If the original origin is ever wanted it belongs in a metadata tag, exactly as suggested, and that can be added later without moving the timestamps back onto the broadcast clock.

One thing is being changed on top, in #577, and it is worth explaining rather than just doing.

The origin is latched from the first packet WRITTEN, which is the arming video keyframe. An MPEG-TS interleaves the audio belonging to a picture ahead of that picture, so the audio packets arriving just after the keyframe carry timestamps EARLIER than the origin, and they are what meets the clamp. Measured here on a source whose audio leads by 100 ms: four AC-3 frames come out stamped at instant zero while the picture sits at zero as well, so the head of the recording loses precisely the offset a rebase is supposed to preserve. The rest of the file is unaffected, it is the first tenth of a second that flattens.

libavformat already has this: `avoid_negative_ts=make_zero` peeks into the interleaving queue for the lowest timestamp across every stream, latches a single offset from it and applies that offset to all streams for the rest of the file. Same result for the case in the report, nothing to clamp, and both invariants from the PR hold by construction rather than by care: `AV_NOPTS_VALUE` travels through, and nothing is reordered.

The tests moved with it and now assert on the written file rather than on the arithmetic: one records a channel whose clock stands at 24549.835 s and reads the first timestamp back as zero, the other records audio leading the keyframe by 100 ms and reads that lead back intact. The second one fails on this PR as merged, which is what made the case worth catching.

Thanks for both this and #573. The recording path and the VOD axis conversion were found the same afternoon from the same stream, which is not a coincidence anyone could have guessed at from the code.

superuser404notfound:
Follow-up is in: #577, merged as `bba7efa8`. The file still begins at zero, the lead of the interleaved audio survives, and the two tests now assert on the written file rather than on the arithmetic. Nothing else about this PR changed.

